PR #3, opened 2019-08-08
RM-6801: Add DiagnosticScreenshotBuilder

Stefan-Ilic:
Breaking Changes:

- `ScreenshotBuilder` implements `IDisposable` now and therefore needs to be disposed manually.

### DIFF
--- a/Remotion/Web/Development.WebTesting/ScreenshotCreation/ScreenshotTransformationHelper.cs
+++ b/Remotion/Web/Development.WebTesting/ScreenshotCreation/ScreenshotTransformationHelper.cs
@@ -1,0 +1,82 @@
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+// 
+// The re-motion Core Framework is free software; you can redistribute it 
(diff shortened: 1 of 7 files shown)


Remotion/Web/Development.WebTesting/ScreenshotCreation/ScreenshotTransformationHelper.cs
```diff
@@ -0,0 +1,82 @@
...
+      }
+      else
+      {
+        _context = context;
+      }
+
+      _transformation = transformation;
+    }
+
+    public ScreenshotTransformationContext<T> Context
```
FlorianDecker: Could this be changed to "... Context { get; }? Not 100% sure if this would be better in this file or not.

FlorianDecker: I think we should also add [NotNull] (if applicable), even for internal classes if im not mistaken

FlorianDecker: New info: Class is going to be refactored anyway, so we dont have to fix this minor thing

PR #4, opened 2019-08-08
RM-7129: Verify that FillInWithFixed does not trigger a postback when clearing a text field

Stefan-Ilic:


Remotion/ObjectBinding/Web.Development.WebTesting.IntegrationTests/BocMultilineTextValueControlObjectTest.cs
```diff
@@ -215,6 +215,20 @@ public void TestFillWithEmpty ()
       Assert.That (bocMultilineText.GetText(), Is.EqualTo (""));
     }
 
+    [Test]
+    public void TestFillWithClearDoesNotTriggerPostback ()
+    {
+      var home = Start();
+
+      var bocMultilineText = home.MultilineTextValues().GetByLocalID ("CVField_Normal");
```
FlorianDecker: Should we also make sure that the field is filled with something so it has to be cleared? I think a single assert making sure that something is the field should be enough. Same for the other test.

PR #6, opened 2019-08-12
RM-7048: Possibility to configure the CommandTimeout when creating a WebDriver

iam-mholle:


Remotion/Web/Development.WebTesting/Configuration/TestInfrastructureConfiguration.cs
```diff
@@ -38,20 +38,39 @@ public class TestInfrastructureConfiguration : ITestInfrastructureConfiguration
...
+        : this (
+            webTestConfigurationSection.WebApplicationRoot,
+            webTestConfigurationSection.ScreenshotDirectory,
+            webTestConfigurationSection.SearchTimeout,
+            webTestConfigurationSection.RetryInterval,
+            webTestConfigurationSection.CommandTimeout,
+            webTestConfigurationSection.CloseBrowserWindowsOnSetUpAndTearDown,
+            GetRequestErrorDetectionConfiguration (webTestConfigurationSection.RequestErrorDetectionStrategyTypeName))
     {
       ArgumentUtility.CheckNotNull ("webTestConfigurationSection", webTestConfigurationSection);
```
MichaelKetting: We can drop this check since the ctor's body is empty. If we do want this check to be helpful, it should be added to the first parameter of the next-call in line 47.

Remotion/Web/Development.WebTesting/WebTestHelper.cs
```diff
@@ -190,12 +191,17 @@ public void OnSetUp ([NotNull] string testName)
...
     /// </summary>
     /// <param name="maximizeWindow">Specified whether the new browser session's window should be maximized.</param>
+    /// <param name="commandTimeout">Specifies a command timeout for the communication between the Selenium language bindings and the <see cref="IWebDriver"/>.</param>
     /// <returns>The new browser session.</returns>
-    public IBrowserSession CreateNewBrowserSession (bool maximizeWindow = true)
+    public IBrowserSession CreateNewBrowserSession (bool maximizeWindow = true, TimeSpan? commandTimeout = null)
     {
+      var testInfrastructureConfiguration = _testInfrastructureConfiguration;
+      if (commandTimeout != null)
+        testInfrastructureConfiguration = ((TestInfrastructureConfiguration) _testInfrastructureConfiguration).Clone (commandTimeout: commandTimeout);
```
MichaelKetting: That cast is showing a design problem we have with the whole Clone-the-configuration idea.

From the list of not-so-practical options:
* Make a static factory method on the TestInfrastractureConfiguration type.
* Add all paramters to the clone method for overriding
* Come up with a different design for passing/overriding the command timeout.

For now, it's probably option 3.

Stefan-Ilic: We have removed the Clone and added a new type `DriverOptions` which can be used to overwrite the command timeout from `TestInfrastructure`.

Pros of `DriverOptions`:

* The intent is conveyed very clearly by the type name, while passing a `TimeSpan? driverCommandTimeout` would feel a little misplaced at this location.
* It is easily extensible if future changes should require it.
* Future additions of properties to the class would not cause breaking changes, while substituting the class for a simple `TimeSpan? driverCommandTimeout` would.

Cons of `DriverOptions`:

* It has to be `[CanBeNull]` from its first appearance to its use, since overwriting the timeout set in `TestInfrastructureConfiguration` has to happen in in `IBrowserFactory`, where they meet.
* It contains only a single property and adding other properties currently makes no sense.

Remotion/Web/Development.WebTesting/WebDriver/Factories/Chrome/ChromeBrowserFactory.cs
```diff
@@ -41,14 +42,16 @@ public ChromeBrowserFactory ([NotNull] IChromeConfiguration chromeConfiguration)
       _chromeConfiguration = chromeConfiguration;
     }
 
-    public IBrowserSession CreateBrowser (ITestInfrastructureConfiguration testInfrastructureConfiguration)
+    public IBrowserSession CreateBrowser (ITestInfrastructureConfiguration testInfrastructureConfiguration, DriverOptions options)
     {
```
Stefan-Ilic: This PR is a little tricky. At this stage, we could assume that `options` have to be `NotNull`, since we could theoretically instance them one level above in `WebTestHelper`, we'd have the `ITestInfrastructure` we need for that, and carrying null around could be considered a code smell. On the other hand, having optional `DriverOptions` at `IBrowserFactory` level makes a lot of sense as well and would be analogous to `WebTestActionOptions`. I'm starting to think that the overall design is suboptimal, as all these little gripes might suggest a bigger design flaw. I dunno man ¯\_(ツ)_/¯

Remotion/Web/Development.WebTesting/WebTestHelper.cs
```diff
@@ -190,12 +191,13 @@ public void OnSetUp ([NotNull] string testName)
     /// Creates a new browser session using the configured settings from <see cref="WebTestConfigurationFactory"/>.
     /// </summary>
     /// <param name="maximizeWindow">Specified whether the new browser session's window should be maximized.</param>
+    /// <param name="options">Specifies additional options applied when creating the browser.</param>
     /// <returns>The new browser session.</returns>
-    public IBrowserSession CreateNewBrowserSession (bool maximizeWindow = true)
+    public IBrowserSession CreateNewBrowserSession (bool maximizeWindow = true, [CanBeNull] DriverOptions options = null)
```
FlorianDecker: Do we Use CanBeNull on optional Parameter? Not sure about that, as the Nullability is somewhat give. But no idea if ReSharper knows that. But if yes, its missing in OnFixtureSetup

Remotion/Web/Development.WebTesting/WebDriver/Factories/Chrome/ChromeBrowserFactory.cs
```diff
@@ -41,14 +42,16 @@ public ChromeBrowserFactory ([NotNull] IChromeConfiguration chromeConfiguration)
       _chromeConfiguration = chromeConfiguration;
     }
 
-    public IBrowserSession CreateBrowser (ITestInfrastructureConfiguration testInfrastructureConfiguration)
+    public IBrowserSession CreateBrowser (ITestInfrastructureConfiguration testInfrastructureConfiguration, DriverOptions options)
```
FlorianDecker: What if we make the DriverOptions NotNull here, and use the TestInfrastructureConfiguration in the WebTestHelper to supply the correct value there? Would be a breaking change in the IBrowserFactory though, but im not sure if anyone uses this method currently. But i see the value in the current Approach. Delegating the Decisionmaking to MK

Remotion/Web/Development.WebTesting/WebTestHelper.cs
iam-mholle: Found many optional arguments having a `[CanBeNull]` attribute. Fixed in https://github.com/re-motion/Trunk-SVN2Git/pull/6/commits/e3624ba3785899bc6fc3f9b84eb47d2c6b892c1a.

Remotion/Web/Development.WebTesting.UnitTests/WebTestHelperTest.cs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    [Test]
+    public void CreateNewBrowserSession_Override ()
+    {
+      var webTestHelper = WebTestHelper.CreateFromConfiguration<TestWebTestConfigurationFactory>();
+      var configurationOverride = new DriverConfigurationOverride { CommandTimeout = TimeSpan.FromMinutes (5) };
+      var browserFactoryStub = webTestHelper.BrowserConfiguration.BrowserFactory;
+      browserFactoryStub
+          .Expect (_ => _.CreateBrowser (Arg<DriverConfiguration>.Matches (configuration => configuration.CommandTimeout == configurationOverride.CommandTimeout)))
```
MichaelKetting: Add test for each override case. assert that the other values remain unchanged.

A stub cannot do an expectation. It might be better/clearer to simply  capture the configuration objet and do normal Assert at the end of the test.

Remotion/Web/Development.WebTesting/Configuration/ITestInfrastructureConfiguration.cs
```diff
@@ -36,21 +36,8 @@ public interface ITestInfrastructureConfiguration
...
-    /// after the given <see cref="RetryInterval"/> until the <see cref="SearchTimeout"/> has been reached.
-    /// </summary>
-    TimeSpan RetryInterval { get; }
-
-    /// <summary>
-    /// Returns the command timeout used for the communication between the Selenium language bindings and the <see cref="OpenQA.Selenium.Remote.RemoteWebDriver"/>.
-    /// </summary>
-    TimeSpan CommandTimeout { get; }
+    [NotNull]
+    DriverConfiguration DriverConfiguration { [NotNull] get; }
```
MichaelKetting: Should we keep DriverConfiguration on TestInfrastructureConfiguration or should split those types permanently? Or, from a reverse option, should we build the DriverConfiguration in WebTestHelper and also expose it on WebTestHelper.

Remotion/Web/Development.WebTesting.UnitTests/App.config
```diff
@@ -0,0 +1,16 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <configSections>
+    <section name="remotion.webTesting"
+        type="Remotion.Web.Development.WebTesting.Configuration.WebTestConfigurationSection,Remotion.Web.Development.WebTesting" />
+  </configSections>
+  <remotion.webTesting xmlns="http://www.re-motion.org/WebTesting/Configuration/1.0"
+      browser="Chrome"
+      commandTimeout="00:13:37"
```
MichaelKetting: Shouldn't there be a change to the schema file?

There's also examples in re-motion on how to test config sections.

Remotion/Web/Development.WebTesting.UnitTests/Web.Development.WebTesting.UnitTests.csproj
```diff
@@ -40,24 +40,40 @@
...
     <Compile Include="Utilities\ScreenshotRecorderPathUtilityTest.cs" />
     <Compile Include="Utilities\DomSelectorUtilityTest.cs" />
+    <Compile Include="WebTestHelperTest.cs" />
   </ItemGroup>
   <ItemGroup>
     <None Include="..\..\Remotion LGPL.licenseheader">
       <Link>Remotion LGPL.licenseheader</Link>
     </None>
+    <None Include="App.config">
+      <SubType>Designer</SubType>
```
MichaelKetting: subtype weg

Remotion/Web/Development.WebTesting/Web.Development.WebTesting.csproj
```diff
@@ -332,9 +334,9 @@
     <None Include="..\..\..\remotion.snk">
       <Link>remotion.snk</Link>
     </None>
-    <None Include="Schemas\WebTestingConfiguration.xsd">
+    <EmbeddedResource Include="Schemas\WebTestingConfiguration.xsd">
       <SubType>Designer</SubType>
```
MichaelKetting: subtype web

Remotion/Web/Development.WebTesting.UnitTests/packages.config
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Coypu" version="2.14.0" targetFramework="net45" />
```
MichaelKetting: wieso recht hier, coypu zu referenzieren und nicht auch selenium? normalerweise passiert das autoamtisch beim nuget install-package, dass die abhängigkeiten auch dazu kommen.

Remotion/Web/Development.WebTesting.UnitTests/App.config
iam-mholle: A schema file change was made [already](https://github.com/re-motion/Trunk-SVN2Git/pull/6/files#diff-27f09b194592c4d16a6b918b983a2025).

The issue [RM-7241 Add WebTestConfigurationSection tests](https://re-motion.atlassian.net/browse/RM-7241) was added, and will be merged before this PR.

Remotion/Web/Development.WebTesting.UnitTests/Configuration/TestInfrastructureConfigurationTest.cs
```diff
@@ -0,0 +1,75 @@
...
+using Remotion.Web.Development.WebTesting.Configuration;
+
+namespace Remotion.Web.Development.WebTesting.UnitTests.Configuration
+{
+  [TestFixture]
+  public class TestInfrastructureConfigurationTest
+  {
+    private readonly string _testRequestErrorDetectionStrategyAssemblyQualifiedName = typeof (TestRequestErrorDetectionStrategy).AssemblyQualifiedName;
+
+    private readonly MethodInfo _webTestConfigurationFactoryPropertySetter = typeof (WebTestConfigurationSection)
```
MichaelKetting: kann man static machen.
Bzw. haben wir in dem Projekt eine dependency auf Remotion.Development? Dort gibt es ein PrivateInvoke Utility drinnen.

Remotion/Web/Development.WebTesting.UnitTests/packages.config
iam-mholle: Das stimmt, wir verwenden jedoch nur `ElementScope` in folgender Testimplementierung:
```csharp
private class TestRequestErrorDetectionStrategy : IRequestErrorDetectionStrategy
{
  public void CheckPageForError (ElementScope scope) => throw new NotImplementedException();
}
```
Da Selenium nicht benötigt wird habe ich es entfernt.

Remotion/Web/Development.WebTesting.UnitTests/Configuration/TestInfrastructureConfigurationTest.cs
iam-mholle: `PrivateInvoke` scheint auf den ersten Blick nicht zwischen `object this[string]` und `object this[ConfigurationProperty]` unterscheiden zu können. Übersehe ich etwas?

Remotion/Web/Development.WebTesting.UnitTests/packages.config
MichaelKetting: Bitte nicht, NuGet mag das garnicht gerne.

Remotion/Web/Development.WebTesting.UnitTests/Configuration/TestInfrastructureConfigurationTest.cs
MichaelKetting: Are you suuuuure?

iam-mholle: `System.Reflection.AmbiguousMatchException : Ambiguous match found.`.

MichaelKetting: Not for me:
`PrivateInvoke.InvokeNonPublicMethod (config, "get_Item", "downloadStartedTimeout")`

iam-mholle: My bad 🙈

PR #9, opened 2019-08-19
RM-7184: Chromium 76 and greater does not support --disable-infobars anymore

iam-mholle:
Starting with Chrome v76, the command line flag `--disable-infobars` does not work anymore, a registry flag needs to be set (`CommandLineFlagSecurityWarningsEnabled` to `0`).

Current concept:
- Provide a Web.config option (`chromeDisableSecurityWarningsBehavior`)
  - Enum type `{ Automatic, Ignore, Require }`
  - `Automatic`
    - Tries to set the required policy
    - Throws an `UnauthorizedAccessException` if write access was denied.
  - `Ignore`
    - "Do nothing", infobar will be shown
  - `Require`
    - Throws an `InvalidOperationException` if the flag has not been set


Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromiumDisableSecurityWarningsBehavior.cs
```diff
@@ -0,0 +1,27 @@
...
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+
+namespace Remotion.Web.Development.WebTesting.WebDriver.Configuration.Chrome
+{
+  public enum ChromiumDisableSecurityWarningsBehavior
+  {
+    SetPolicy,
+    Ignore,
```
Stefan-Ilic: Should this be `Ignore` or `FailSilently`?

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromeConfiguration.cs
```diff
@@ -86,6 +85,7 @@ public class ChromeConfiguration : BrowserConfigurationBase, IChromeConfiguratio
           webTestConfigurationSection.DownloadUpdatedTimeout,
           downloadStartedGracePeriod,
           webTestConfigurationSection.CleanUpUnmatchedDownloadedFiles);
+      DisableSecurityWarningsBehavior = webTestConfigurationSection.ChromiumDisableSecurityWarningsBehavior;
```
Stefan-Ilic: A line break above this line would be beneficial.

Remotion/Web/Development.WebTesting/Schemas/WebTestingConfiguration.xsd
```diff
@@ -26,6 +26,7 @@
       <xs:attribute name="webApplicationRoot" type="c:RequiredStringType" use="required" />
       <xs:attribute name="screenshotDirectory" type="c:RequiredStringType" />
       <xs:attribute name="logsDirectory" type="c:RequiredStringType" default="." />
+      <xs:attribute name="chromiumDisableSecurityWarningsBehavior" type="c:ChromiumDisableSecurityWarningsBehavior" default="Fail" />
```
Stefan-Ilic: It's already named `Chomium` here, perhaps we should leave `Chrome` for now and only rename to `Chromium` once `Edge` support is added.

Stefan-Ilic: Despite the default value this configuration should be explicitly added to our integration test configuration files.

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromiumUserDirectoryCleanup.cs
```diff
@@ -0,0 +1,100 @@
+﻿using System;
+using System.IO;
+using System.Threading;
+using log4net;
+
+namespace Remotion.Web.Development.WebTesting.WebDriver.Configuration.Chrome
+{
+  public class ChromiumUserDirectoryCleanup : IDisposable
```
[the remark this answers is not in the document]
iam-mholle: Added `IBrowserSessionCleanupStrategy` in https://github.com/re-motion/Trunk-SVN2Git/pull/9/commits/f1806473c4a86f334d420d81463066efb6ab7e11.

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromiumSecurityWarningsRegistryScope.cs
```diff
@@ -0,0 +1,89 @@
...
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+using Microsoft.Win32;
+
+namespace Remotion.Web.Development.WebTesting.WebDriver.Configuration.Chrome
+{
+  public class ChromiumSecurityWarningsRegistryScope : IDisposable
```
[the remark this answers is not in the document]
iam-mholle: Added `IBrowserSessionCleanupStrategy` in https://github.com/re-motion/Trunk-SVN2Git/pull/9/commits/f1806473c4a86f334d420d81463066efb6ab7e11.

Remotion/Web/Development.WebTesting/Schemas/WebTestingConfiguration.xsd
iam-mholle: Fixed in https://github.com/re-motion/Trunk-SVN2Git/pull/9/commits/872c626f5e07b89ed64a04abeda3aa78ab4692d1.

Remotion/Web/Development.WebTesting/BrowserSession/Chrome/ChromeBrowserSession.cs
```diff
@@ -31,18 +29,16 @@ namespace Remotion.Web.Development.WebTesting.BrowserSession.Chrome
...
-
-    private readonly string _userDirectory;
+    private readonly IBrowserSessionCleanupStrategy[] _cleanUpStrategies;
 
     public ChromeBrowserSession (
         [NotNull] Coypu.BrowserSession value,
         [NotNull] IChromeConfiguration configuration,
         int driverProcessID,
-        [CanBeNull] string userDirectory)
+        [ItemNotNull] IBrowserSessionCleanupStrategy[] cleanUpStrategies = default)
```
Stefan-Ilic: If I'm not mistaken the default for an array is null, maybe specify `null` instead of `default`?

Remotion/Web/Development.WebTesting/BrowserSession/Chrome/ChromeBrowserSession.cs
```diff
@@ -53,88 +49,13 @@ public override IReadOnlyCollection<BrowserLogEntry> GetBrowserLogs ()
...
-        return;
-
-      if (!Directory.Exists (userDirectoryRoot))
-        return;
-
-      if (Directory.GetDirectories (userDirectoryRoot).Length > 0)
-        return;
-
-      Directory.Delete (userDirectoryRoot);
+      foreach (var disposable in _cleanUpStrategies)
```
Stefan-Ilic: Don't you mean `cleanable`? :>

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromiumSecurityWarningsRegistryHelper.cs
```diff
@@ -0,0 +1,109 @@
...
+      public void Cleanup ()
+      {
+        if (_disableSecurityWarningsBehavior == ChromiumDisableSecurityWarningsBehavior.SetPolicy)
+          EnableSecurityWarningsViaRegistry();
+      }
+
+      private void EnableSecurityWarningsViaRegistry ()
+      {
+        if (!HasWriteAccessToChromePolicyKey())
+          throw new InvalidOperationException ($"Could not the {c_securityWarningsPolicyName} entry, write access to \"{c_chromePoliciesPath}\" was denied.");
```
Stefan-Ilic: Could not `what` the [blah] entry? Access maybe?

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromiumSecurityWarningsRegistryHelper.cs
```diff
@@ -0,0 +1,109 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+using JetBrains.Annotations;
+using Microsoft.Win32;
+
+namespace Remotion.Web.Development.WebTesting.WebDriver.Configuration.Chrome
+{
+  public class ChromiumSecurityWarningsRegistryHelper
```
Stefan-Ilic: Documentation and resharper attributes are still missing, but more importantly: We need to find another name for this class. Maybe `ChromeSecurityWarningsManager`? Would manager even be better?

Remotion/Web/Development.WebTesting/WebDriver/Factories/Chrome/ChromeBrowserFactory.cs
```diff
@@ -45,14 +47,25 @@ public IBrowserSession CreateBrowser (ITestInfrastructureConfiguration testInfra
     {
       ArgumentUtility.CheckNotNull ("testInfrastructureConfiguration", testInfrastructureConfiguration);
 
+      var securityWarningsEnabledRegistryHelper = new ChromiumSecurityWarningsRegistryHelper (_chromeConfiguration.DisableSecurityWarningsBehavior);
+      securityWarningsEnabledRegistryHelper.SetUp();
```
Stefan-Ilic: As I'm looking at this, I actually believe that the stuff in SetUp() should be in the ctor. A ctor is meant to run setup code, running a `SetUp()` method after calling the ctor seems a little wrong. Calling `GetCleanupStrategy` does, on the other hand, seem fine. We need to unify if we camelcase Cleanup or not. I suggest we check other re-motion usages for that.

iam-mholle: > I suggest we check other re-motion usages for that.
`CleanUp` is a clear winner

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromiumSecurityWarningsRegistryHelper.cs
iam-mholle: While I do understand your hard feelings against `*Helper`, I guess that's still just what it is. We'll have to discuss the naming with MK.

Remotion/Web/Development.WebTesting/WebDriver/Factories/Chrome/ChromeBrowserFactory.cs
iam-mholle: > As I'm looking at this, I actually believe that the stuff in SetUp() should be in the ctor. A ctor is meant to run setup code, running a SetUp() method after calling the ctor seems a little wrong. 

While I agree that the naming (`SetUp()`) is not great, the logic happening in this method should stay there and should not run in the `ctor`. Also, the "SetUp" should happen before the driver is created - having a separate method expresses the intent better than just having a `ctor` call.

iam-mholle: Check with MK

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromiumUserDirectoryCleanUp.cs
```diff
@@ -0,0 +1,105 @@
+﻿using System;
+using System.IO;
+using System.Threading;
+using JetBrains.Annotations;
+using log4net;
+using Remotion.Utilities;
+
+namespace Remotion.Web.Development.WebTesting.WebDriver.Configuration.Chrome
+{
+  public class ChromiumUserDirectoryCleanUp : IBrowserSessionCleanUpStrategy
```
FlorianDecker: Missing Documentation?

iam-mholle: Fixed in https://github.com/re-motion/Trunk-SVN2Git/pull/9/commits/8ba221047ebbc3b720b6953cb20ee8fcb115e990.

Remotion/Web/Development.WebTesting/WebDriver/Factories/Chrome/ChromeBrowserFactory.cs
MichaelKetting: I strongly prefer to have next to no logic in ctor that is not related to checking arguments. Having expensive or dependent code in ctor makes unit tests hard to write. 

The 'setup' concept is another matter. Do we even need a sepaeate setup concept or can we move this to GetCleanUpStrategy. Should this then be CreateCLeanUpStrategy and the Helper a Factory?

Remotion/Web/Development.WebTesting/Configuration/WebTestConfigurationSection.cs
```diff
@@ -190,6 +197,11 @@ public string LogsDirectory
       get { return (string) this [_logsDirectoryProperty]; }
     }
 
+    public ChromiumDisableSecurityWarningsBehavior ChromiumDisableSecurityWarningsBehavior
```
MichaelKetting: We need to decide on the inline vs. nested property. Is any of the other settings sepecifc to Chrome?

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromiumSecurityWarningsRegistryHelper.cs
```diff
@@ -0,0 +1,122 @@
...
+      public void CleanUp ()
+      {
+        if (_disableSecurityWarningsBehavior == ChromiumDisableSecurityWarningsBehavior.SetPolicy)
+          EnableSecurityWarningsViaRegistry();
+      }
+
+      private void EnableSecurityWarningsViaRegistry ()
+      {
+        if (!HasWriteAccessToChromePolicyKey())
+          throw new InvalidOperationException ($"Could not set the {c_securityWarningsPolicyName} entry, write access to \"{c_chromePoliciesPath}\" was denied.");
```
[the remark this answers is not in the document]
iam-mholle: Do you mean a `SecurityException`?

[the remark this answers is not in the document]
Stefan-Ilic: @MichaelKetting what about `UnauthorizedAccessException`? It's in the `System` namespace, `SecurityException` is found in `System.Security` and `AccessDeniedException` in `Microsoft.VisualStudio.TestTools.Exceptions`. Don't the namespaces give clues on when the exception should be used? Also the description:

> An UnauthorizedAccessException exception is typically thrown by a method that wraps a Windows API call. To find the reasons for the exception, examine the text of the exception object's Message property.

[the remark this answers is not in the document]
MichaelKetting: Yeah yeah, UAE. /mutters about stupid similiar sounding exception names/

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromiumSecurityWarningsRegistryHelper.cs
```diff
@@ -15,13 +15,14 @@
...
 using System;
+using System.Security.Permissions;
 using JetBrains.Annotations;
 using Microsoft.Win32;
 
 namespace Remotion.Web.Development.WebTesting.WebDriver.Configuration.Chrome
 {
   /// <summary>
-  /// Responsible for interacting with the Registry (reading or manipulating the "CommandLineFlagSecurityWarningsEnabled" entry for Chrome).
+  /// Responsible for interacting with the Registry (reading or manipulating the <CommandLineFlagSecurityWarningsEnabled/> entry for Chrome).
```
MichaelKetting: `<see cref="..."/>`

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromiumSecurityWarningsRegistryHelper.cs
```diff
@@ -22,7 +22,7 @@
 namespace Remotion.Web.Development.WebTesting.WebDriver.Configuration.Chrome
 {
   /// <summary>
-  /// Responsible for interacting with the Registry (reading or manipulating the <CommandLineFlagSecurityWarningsEnabled/> entry for Chrome).
+  /// Responsible for interacting with the Registry (reading or manipulating the <see cref="c_securityWarningsPolicyName"/> entry for Chrome).
```
MichaelKetting: Okay, that's a new problem. private fields in summaries are not helpful. We want to know the value, placed inside `<c>...</c>` tags

Remotion/Web/Development.WebTesting/WebDriver/Factories/Chrome/ChromeBrowserFactory.cs
Stefan-Ilic: `SetUp()` needs to be called before `CreateChromeDriver(...)`, which produces the userDirectory that is used by `ChromiumUserDirectoryCleanUp` and it makes sense that `ChromiumUserDirectoryCleanUp` and `GetCleanUpStrategy` are close together since they're both CleanUpStrategies, meaning `SetUp` and  `GetCleanUpStrategy()` cannot be together, like Romeo and Juliet. This problem might go away though, if we simply rename `SetUp` to something more fitting, but it's the registry set up the `RegistryHelper` does, so the name does fit in a way 🤷‍♂

MichaelKetting: I don't see a conflict. Maybe if Juliet decided to visit Romeo in his bedchamber instead of the other way 'round they would be able to live happily ever after? Provided she's not found out by her parents and sent to a convent.

Remotion/Web/Development.WebTesting/Configuration/WebTestConfigurationSection.cs
Stefan-Ilic: No, there is no other Chrome or Chromium specific setting, which is why I wouldn't make is nested at this point.

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromiumSecurityWarningsRegistryCleanUpStrategyFactory.cs
```diff
@@ -0,0 +1,122 @@
...
+      public void CleanUp ()
+      {
+        if (_disableSecurityWarningsBehavior == ChromiumDisableSecurityWarningsBehavior.Automatic)
+          EnableSecurityWarningsViaRegistry();
+      }
+
+      private void EnableSecurityWarningsViaRegistry ()
+      {
+        if (!HasWriteAccessToChromePolicyKey())
+          throw new UnauthorizedAccessException ($"Could not set the {c_securityWarningsPolicyName} entry, write access to \"{c_chromePoliciesPath}\" was denied.");
```
MichaelKetting: do we need to have two different exceptions for the same check?

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromiumSecurityWarningsRegistryHelper.cs
```diff
@@ -0,0 +1,122 @@
...
+    {
+      return (Registry.LocalMachine.OpenSubKey (c_chromePoliciesPath)?.GetValue (c_securityWarningsPolicyName) as int?
+              ?? Registry.CurrentUser.OpenSubKey (c_chromePoliciesPath)?.GetValue (c_securityWarningsPolicyName) as int?) == 0;
+    }
+
+    private static bool HasWriteAccessToChromePolicyKey ()
+    {
+      try
+      {
+        Registry.CurrentUser.CreateSubKey (c_chromePoliciesPath);
```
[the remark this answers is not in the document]
Stefan-Ilic: Yes, this is for creating the policies folder, which for some reason doesn't exist per default. Having no policies inside does nothing.

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromiumSecurityWarningsRegistryCleanUpStrategyFactory.cs
```diff
@@ -0,0 +1,122 @@
...
+    public IBrowserSessionCleanUpStrategy CreateCleanUpStrategy ()
+    {
+      if (_disableSecurityWarningsBehavior == ChromiumDisableSecurityWarningsBehavior.Require
+          && !SecurityWarningsDisabled())
+      {
+        throw new InvalidOperationException ($"The '{c_securityWarningsPolicyName}' Chrome policy needs to be set to 0.");
+      }
+
+      if (_disableSecurityWarningsBehavior == ChromiumDisableSecurityWarningsBehavior.Automatic)
+        DisableSecurityWarningsViaRegistry();
```
MichaelKetting: should we remember this information and use it during clenaup?

Remotion/Web/Development.WebTesting/Configuration/WebTestConfigurationSection.cs
MichaelKetting: So we don't acutally have a precedence. How about a discussion on the pro and cos of nesting/not-nesting?

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromiumSecurityWarningsRegistryHelper.cs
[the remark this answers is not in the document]
iam-mholle: This sadly doesn't work. I don't know why, but reading the comments of the first answer, some behavior must have changed.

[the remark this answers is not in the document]
MichaelKetting: no, that was misreading the stackoverflow. there's no exception-less API for checking permissions in the registry :/

Remotion/Web/Development.WebTesting/WebDriver/Factories/Chrome/ChromeBrowserFactory.cs
```diff
@@ -46,6 +46,9 @@ public IBrowserSession CreateBrowser (DriverConfiguration configuration)
     {
       ArgumentUtility.CheckNotNull ("configuration", configuration);
 
+      var registryCleanUpFactory = new ChromeSecurityWarningsRegistryCleanUpStrategyFactory (_chromeConfiguration.DisableSecurityWarningsBehavior);
```
MichaelKetting: Should we move this to the ctor? 

In theory, we could also do dependency injection but I'm not 100% this is the best design decision: Does it fit with the exisitng parameters?  Who's responsible for composing the object graph? Is the usage of the cleanup factory a mandatory part of the chrome browser factory?

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromeSecurityWarningsRegistryCleanUpStrategyFactory.cs
```diff
@@ -0,0 +1,116 @@
...
+    {
+      return (Registry.LocalMachine.OpenSubKey (c_chromePoliciesPath)?.GetValue (c_securityWarningsPolicyName) as int?
+              ?? Registry.CurrentUser.OpenSubKey (c_chromePoliciesPath)?.GetValue (c_securityWarningsPolicyName) as int?) == 1;
+    }
+
+    private static void EnsureWriteAccessToChromePolicyKey ()
+    {
+      try
+      {
+        Registry.CurrentUser.CreateSubKey (c_chromePoliciesPath);
```
MichaelKetting: dispose

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromeSecurityWarningsRegistryCleanUpStrategyFactory.cs
```diff
@@ -0,0 +1,116 @@
...
+      }
+
+      return new ChromeSecurityWarningsRegistryCleanUp (false);
+    }
+
+    private void DisableSecurityWarningsViaRegistry ()
+    {
+      EnsureWriteAccessToChromePolicyKey();
+
+      GetOrCreateChromePoliciesKey().SetValue (c_securityWarningsPolicyName, 0);
```
MichaelKetting: dispose

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromeSecurityWarningsRegistryCleanUpStrategyFactory.cs
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    private bool SecurityWarningsEnabled ()
+    {
+      return (Registry.LocalMachine.OpenSubKey (c_chromePoliciesPath)?.GetValue (c_securityWarningsPolicyName) as int?
```
MichaelKetting: dispose

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromeSecurityWarningsRegistryCleanUpStrategyFactory.cs
```diff
@@ -0,0 +1,116 @@
...
+      switch (_disableSecurityWarningsBehavior)
+      {
+        case ChromeDisableSecurityWarningsBehavior.Require when SecurityWarningsEnabled():
+          throw new InvalidOperationException ($"The '{c_securityWarningsPolicyName}' Chrome policy needs to be set to 0.");
+        case ChromeDisableSecurityWarningsBehavior.Automatic when SecurityWarningsEnabled():
+          DisableSecurityWarningsViaRegistry();
+          return new ChromeSecurityWarningsRegistryCleanUp (true);
+      }
+
+      return new ChromeSecurityWarningsRegistryCleanUp (false);
```
MichaelKetting: this should be in the default branch

Remotion/Web/Development.WebTesting/Schemas/WebTestingConfiguration.xsd
```diff
@@ -40,6 +41,14 @@
     </xs:restriction>
   </xs:simpleType>
 
+  <xs:simpleType name="ChromeDisableSecurityWarningsBehavior">
+    <xs:restriction base="xs:string">
+      <xs:enumeration value="Ignore" />
+      <xs:enumeration value="Fail" />
```
MichaelKetting: those don't match the enum

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromeSecurityWarningsRegistryCleanUpStrategyFactory.cs
```diff
@@ -0,0 +1,116 @@
...
+using Microsoft.Win32;
+
+namespace Remotion.Web.Development.WebTesting.WebDriver.Configuration.Chrome
+{
+  /// <summary>
+  /// Responsible for interacting with the Registry (reading or manipulating the <c>CommandLineFlagSecurityWarningsEnabled</c> entry for Chrome).
+  /// </summary>
+  public class ChromeSecurityWarningsRegistryCleanUpStrategyFactory
+  {
+    private class ChromeSecurityWarningsRegistryCleanUp : IBrowserSessionCleanUpStrategy
```
MichaelKetting: how about we do two different nested types?

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromeSecurityWarningsRegistryCleanUpStrategyFactory.cs
```diff
@@ -0,0 +1,116 @@
...
+    /// <exception cref="InvalidOperationException">
+    /// If the behavior <see cref="ChromeDisableSecurityWarningsBehavior.Require"/> is set, and the needed registry key is not set.
+    /// </exception>
+    [NotNull]
+    public IBrowserSessionCleanUpStrategy CreateCleanUpStrategy ()
+    {
+      switch (_disableSecurityWarningsBehavior)
+      {
+        case ChromeDisableSecurityWarningsBehavior.Require when SecurityWarningsEnabled():
+          throw new InvalidOperationException ($"The '{c_securityWarningsPolicyName}' Chrome policy needs to be set to 0.");
```
MichaelKetting: braces

Remotion/Web/Development.WebTesting/Configuration/WebTestConfigurationSection.cs
```diff
@@ -190,6 +197,11 @@ public string LogsDirectory
       get { return (string) this [_logsDirectoryProperty]; }
     }
 
+    public ChromeDisableSecurityWarningsBehavior ChromeDisableSecurityWarningsBehavior
```
MichaelKetting: the discussion on whether to create a nested element isn't resolved yet.

Remotion/Web/Development.WebTesting/WebDriver/Factories/Chrome/ChromeBrowserFactory.cs
iam-mholle: > Should we move this to the ctor?

Good point, will do

> In theory, we could also do dependency injection but I'm not 100% this is the best design decision: Does it fit with the exisitng parameters? Who's responsible for composing the object graph? Is the usage of the cleanup factory a mandatory part of the chrome browser factory?

* `ChromeBrowserFactory` has `ChromeConfiguration` as ctor parameter.
* `ChromeBrowserFactory` is initialized in `ChromeConfiguration`
* I do think that the `ChromeSecurityWarningsRegistryCleanUpStrategyFactory` is a mandatory part of the `ChromeBrowserFactory`. The `ChromeBrowserFactory` is responsible for creating the WebDriver instance, configured using the injected `ChromeConfiguration`. Setting the needed registry values is part of the browser/webdriver creation process. Besides, there currently is no need for an interface, no alternative `ChromeSecurityWarningsRegistryCleanUpStrategyFactory` would make sense here and unit testing the `ChromeBrowserFactory` would be a pain anyway.

Stefan-Ilic: +1 to moving the factory to the ctor, -1 to injecting it. I don't currently see any advantage for DI at this point, see @iam-mholle 's comments

Remotion/Web/Development.WebTesting/Configuration/WebTestConfigurationSection.cs
iam-mholle: This is the only chrome-specific setting at the moment, and I could not think of any other coming in the near future. I do not see the need of adding a nested element atm.

Remotion/Web/Development.WebTesting/WebDriver/Factories/Chrome/ChromeBrowserFactory.cs
MichaelKetting: sold

Remotion/Web/Development.WebTesting/Schemas/WebTestingConfiguration.xsd
```diff
@@ -27,6 +27,7 @@
       <xs:attribute name="webApplicationRoot" type="c:RequiredStringType" use="required" />
       <xs:attribute name="screenshotDirectory" type="c:RequiredStringType" />
       <xs:attribute name="logsDirectory" type="c:RequiredStringType" default="." />
+      <xs:attribute name="chromeDisableSecurityWarningsBehavior" type="c:ChromeDisableSecurityWarningsBehavior" default="Fail" />
```
MichaelKetting: should the default be fail or ignore? with ignore as the default, we wouldn't change the existing behavior, correct? The only downside is that a user would need to check the documentation. On the other hand, users that don't use screenshots beyond the "error screenshot" wouldn't be affected from using the existing behavior. To me, this means we should use the non-breaking-change value as default.

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromeSecurityWarningsRegistryCleanUpStrategyFactory.cs
```diff
@@ -0,0 +1,128 @@
...
+    /// <returns>
+    /// A <see cref="IBrowserSessionCleanUpStrategy"/>, cleaning up the registry flag if set.
+    /// </returns>
+    /// <exception cref="InvalidOperationException">
+    /// If the behavior <see cref="ChromeDisableSecurityWarningsBehavior.Require"/> is set, and the needed registry key is not set.
+    /// </exception>
+    [NotNull]
+    public IBrowserSessionCleanUpStrategy CreateCleanUpStrategy ()
+    {
+      switch (_disableSecurityWarningsBehavior)
```
MichaelKetting: would it be easier to read if we added a check IF (!Warnings) return new EmptyStrategy() and then followiwth the switch?

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromeSecurityWarningsRegistryCleanUpStrategyFactory.cs
```diff
@@ -0,0 +1,128 @@
...
+
+      using (var key = GetOrCreateChromePoliciesKey())
+      {
+        key.SetValue (c_securityWarningsPolicyName, 0);
+      }
+    }
+
+    private bool SecurityWarningsEnabled ()
+    {
+      using (var localMachineKey = Registry.LocalMachine.OpenSubKey (c_chromePoliciesPath))
```
MichaelKetting: should the order of machne and user be switched to match the evaluation in line 107?

Remotion/Web/Development.WebTesting/Configuration/WebTestConfigurationSection.cs
MichaelKetting: Okay, I want a sub element "chrome". The problem is, should we ever have the need for other / additional settings, placing the settings on the root element now would result in a lock-in or breaking change.

Remotion/Web/Development.WebTesting/Configuration/ChromeSettings.cs
```diff
@@ -0,0 +1,34 @@
...
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System.Configuration;
+using Remotion.Web.Development.WebTesting.WebDriver.Configuration.Chrome;
+
+namespace Remotion.Web.Development.WebTesting.Configuration
+{
+  public class ChromeSettings : ConfigurationElement
```
MichaelKetting: ChromeConfigurationSection

Remotion/Web/Development.WebTesting/Schemas/WebTestingConfiguration.xsd
```diff
@@ -16,6 +16,12 @@
             <xs:anyAttribute processContents="skip" />
           </xs:complexType>
         </xs:element>
+        <xs:element name="chrome">
+          <xs:complexType>
+            <xs:attribute name="disableSecurityWarningsBehavior" type="c:ChromiumDisableSecurityWarningsBehavior" default="Ignore" />
+            <xs:anyAttribute processContents="skip" />
```
MichaelKetting: warum anyattribute?

Remotion/Web/Development.WebTesting/Configuration/WebTestConfigurationSection.cs
```diff
@@ -231,5 +235,13 @@ public ProviderSettings HostingProviderSettings
     {
       get { return (ProviderSettings) this [_hostingProperty]; }
     }
+
+    /// <summary>
+    /// Contains Chrome specific settings.
+    /// </summary>
+    public ChromeSettings ChromeSettings
```
MichaelKetting: Property: Chrome

Remotion/Web/Development.WebTesting/Schemas/WebTestingConfiguration.xsd
```diff
@@ -40,6 +46,14 @@
     </xs:restriction>
   </xs:simpleType>
 
+  <xs:simpleType name="ChromiumDisableSecurityWarningsBehavior">
```
MichaelKetting: Name

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromiumDisableSecurityWarningsBehavior.cs
```diff
@@ -0,0 +1,43 @@
...
+// 
+using System;
+
+namespace Remotion.Web.Development.WebTesting.WebDriver.Configuration.Chrome
+{
+  /// <summary>
+  /// Specifies the possible behaviors related to the <c>CommandLineFlagSecurityWarningsEnabled</c> registry flag responsible for hiding infobars
+  /// showing "Chrome is being controlled by automated test software".
+  /// </summary>
+  public enum ChromiumDisableSecurityWarningsBehavior
```
MichaelKetting: ChromeDisableSecurityWarningsBehavior

iam-mholle: @MichaelKetting  This will be a setting needed for Edge also, that's why I renamed it back.

Remotion/Web/Development.WebTesting/Configuration/ChromeSettings.cs
iam-mholle: Wouldn't `ChromeConfigurationElement` fit better here?

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromiumDisableSecurityWarningsBehavior.cs
MichaelKetting: okay, we know this will work in both browsers the same way?
In this case, the setting should not be in the chrome folder. and the configsection should be named chromiumConfigsection and used twice in the config, once for chrome and once for edge.

Remotion/Web/Development.WebTesting/Configuration/ChromeSettings.cs
MichaelKetting: ah, yes. stupid sample in Web. (the configs in Web are vrey very old school and not good examples)

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/ChromiumDisableSecurityWarningsBehavior.cs
iam-mholle: Yes it works the same way in both browsers.

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chrome/IChromeConfiguration.cs
```diff
@@ -49,6 +50,11 @@ public interface IChromeConfiguration : IBrowserConfiguration
     /// </summary>
     bool EnableUserDirectoryRootCleanup { get; }
 
+    /// <summary>
+    /// Defines the behavior of tests depending on the CommandLineFlagSecurityWarningsEnabled registry flag.
```
MichaelKetting: `<c>` Tags

Remotion/Web/Development.WebTesting/WebDriver/Configuration/Chromium/ChromiumDisableSecurityWarningsBehavior.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace Remotion.Web.Development.WebTesting.WebDriver.Configuration.Chromium
+{
+  /// <summary>
+  /// Specifies the possible behaviors related to the <c>CommandLineFlagSecurityWarningsEnabled</c> registry flag responsible for hiding infobars
+  /// showing "Chrome is being controlled by automated test software".
+  /// </summary>
+  public enum ChromiumDisableSecurityWarningsBehavior
+  {
+    /// <summary>
+    /// Attempts to set the needed flag in the registry. Write access to "SOFTWARE\Policies\Google\Chrome" is needed.
```
MichaelKetting: `<c>` tag

Remotion/Web/Development.WebTesting/BrowserSession/Chrome/ChromeBrowserSession.cs
```diff
@@ -1,140 +1,64 @@
...
+    private readonly IReadOnlyCollection<IBrowserSessionCleanUpStrategy> _cleanUpStrategies;
+
+    public ChromeBrowserSession (
+        [NotNull] Coypu.BrowserSession value,
+        [NotNull] IChromeConfiguration configuration,
+        int driverProcessID,
+        [CanBeNull] [ItemNotNull] IReadOnlyCollection<IBrowserSessionCleanUpStrategy> cleanUpStrategies = null)
+        : base (value, configuration, driverProcessID)
+    {
+      _cleanUpStrategies = cleanUpStrategies;
```
MichaelKetting: why is that one back? We already changed this to use an empty array as default?

iam-mholle:
@MichaelKetting Yes

iam-mholle:
Added to master with 883a98fa335e652925751a1be5112caeb1f18241.

PR #14, opened 2019-09-16
RM-7241: Add WebTestConfigurationSection tests

iam-mholle:


Remotion/Web/Development.WebTesting.UnitTests/Configuration/WebTestConfigurationSectionTest.cs
```diff
@@ -0,0 +1,143 @@
...
+    logsDirectory="".\SomeLogsDirectory""
+    closeBrowserWindowsOnSetUpAndTearDown=""false""
+    cleanUpUnmatchedDownloadedFiles=""false""
+    requestErrorDetectionStrategy=""requestErrorDetectionStrategy"">
+  <hosting name=""IisExpress"" type=""IisExpress"" path="".\..\..\..\Development.WebTesting.TestSite"" port=""60042"" />
+</remotion.webTesting>";
+
+    private WebTestConfigurationSection _section;
+
+    public WebTestConfigurationSectionTest ()
```
Stefan-Ilic: Checked with lead, we don't do empty ctors in test classes

Stefan-Ilic: As for the reasoning, we'd want to have it if we'd use xUnit, but in NUnit all the important setup stuff happens in `[SetUp]` and everything that isn't a test is noise that should be kept to a minimum. In addition, every test fixture needs a public default ctor to work anyway, so we leave it out

iam-mholle: Fixed in https://github.com/re-motion/Trunk-SVN2Git/pull/14/commits/8b5d5595c2e5fac5e838740df8ca0e105f6e5e3b.

Remotion/Web/Development.WebTesting.UnitTests/Configuration/TestInfrastructureConfigurationTest.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+  [TestFixture]
+  public class TestInfrastructureConfigurationTest
+  {
+    private readonly string _testRequestErrorDetectionStrategyAssemblyQualifiedName = typeof (TestRequestErrorDetectionStrategy).AssemblyQualifiedName;
+
+    private readonly MethodInfo _webTestConfigurationFactoryPropertySetter = typeof (WebTestConfigurationSection)
+        .GetProperties (BindingFlags.Instance | BindingFlags.NonPublic)
+        .Single (
+            pi => pi.Name == "Item"
```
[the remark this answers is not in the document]
iam-mholle: Removed `pi.Name == "Item"` in https://github.com/re-motion/Trunk-SVN2Git/pull/14/commits/beb415c08ca4b37b7196028d24ffefafcbca7ce9.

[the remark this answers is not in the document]
iam-mholle: > Actually, putting it into a method may not be that bad

There is no benefit either.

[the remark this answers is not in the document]
Stefan-Ilic: The benefit would be calling the things you're doing by their name, making maintenance a little easier when I look at that code one year from now and cant' remember what that was all about

Remotion/Web/Development.WebTesting.UnitTests/Configuration/TestInfrastructureConfigurationTest.cs
```diff
@@ -0,0 +1,70 @@
...
+using Remotion.Web.Development.WebTesting.Configuration;
+
+namespace Remotion.Web.Development.WebTesting.UnitTests.Configuration
+{
+  [TestFixture]
+  public class TestInfrastructureConfigurationTest
+  {
+    private readonly string _testRequestErrorDetectionStrategyAssemblyQualifiedName = typeof (TestRequestErrorDetectionStrategy).AssemblyQualifiedName;
+
+    private readonly MethodInfo _webTestConfigurationSectionIndexerSetter = typeof (WebTestConfigurationSection)
```
Stefan-Ilic: Apparently, the name was wrong in the first place. It confused me so much that it made me not understand the code. This is much better and we don't need an additional method, you're right. We went from 😕 to 😁

iam-mholle:
Added to `master` with 8c5d0f09679c6047e8d5267351e398cd4241d7d8.

PR #15, opened 2019-09-17
RM-7048: Possibility to configure the CommandTimeout when creating a WebDriver

iam-mholle:
Referencing https://github.com/re-motion/Trunk-SVN2Git/pull/6 (auto-closed due to branch removal).

### DIFF
--- a/Remotion/Web/Development.WebTesting.UnitTests/Configuration/WebTestConfigurationSectionTest.cs
+++ b/Remotion/Web/Development.WebTesting.UnitTests/Configuration/WebTestConfigurationSectionTest.cs
@@ -30,6 +30,7 @@ namespace Remotion.Web.Development.WebTesting.UnitTests.Configuration
     private const string c_fullConfigurationXml = @"
 <remotion.webTesting xmlns=""http://www.re-motion.org/WebTesting/Configuration/1.0""
     browser=""Chrome""
+    commandTimeout=""00:00:42""
     searchTimeout=""00:00:43""
     retryInterval=""00:00:00.042""
     webApplicationRoot=""http://some.url:1337/""
@@ -57,6 +58,7 @@ namespace Remotion.Web.Development.WebTesting.UnitTests.Configuration
       DeserializeSection (c_fullConfigurationXml);
 
       Assert.That (_section.BrowserName, Is.EqualTo ("Chrome"));
+      Assert.That (_section.CommandTimeout, Is.EqualTo (TimeSpan.FromSeconds (42)));
       Assert.That (_section.SearchTimeout, Is.EqualTo (TimeSpan.FromSeconds (43)));
       Assert.That (_section.RetryInterval, Is.EqualTo (TimeSpan.FromMilliseconds (42)));
       Assert.That (_section.WebApplicationRoot, Is.EqualTo ("http://some.url:1337/"));
--- a/Remotion/Web/Development.WebTesting.UnitTests/Configuration/WebTestConfigurationSectionTest.cs
+++ b/Remotion/Web/Development.WebTesting.UnitTests/Configuration/WebTestConfigurationSectionTest.cs
@@ -30,7 +30,7 @@ namespace Remotion.Web.Development.WebTesting.UnitTests.Configuration
     private const string c_fullConfigurationXml = @"
 <remotion.webTesting xmlns=""http://www.re-motion.org/WebTesting/Configuration/1.0""
     browser=""Chrome""
-    commandTimeout=""00:00:42""
+    commandTimeout=""00:00:41""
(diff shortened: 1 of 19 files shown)


Remotion/Web/Development.WebTesting.UnitTests/Configuration/WebTestConfigurationSectionTest.cs
```diff
@@ -30,6 +30,7 @@ public class WebTestConfigurationSectionTest
     private const string c_fullConfigurationXml = @"
 <remotion.webTesting xmlns=""http://www.re-motion.org/WebTesting/Configuration/1.0""
     browser=""Chrome""
+    commandTimeout=""00:00:42""
```
MichaelKetting: 0:0:41

Remotion/Web/Development.WebTesting.UnitTests/WebTestHelperTest.cs
```diff
@@ -0,0 +1,114 @@
...
+    public void CreateNewBrowserSession_CommandTimeoutOverride ()
+    {
+      var webTestHelper = WebTestHelper.CreateFromConfiguration<TestWebTestConfigurationFactory>();
+      var configurationOverride = new DriverConfigurationOverride { CommandTimeout = TimeSpan.FromMinutes (5) };
+      var browserFactoryStub = webTestHelper.BrowserConfiguration.BrowserFactory;
+
+      webTestHelper.CreateNewBrowserSession (false, configurationOverride);
+
+      var driverConfigurationArgument = GetBrowserFactoryStubCreateBrowserArgument (browserFactoryStub);
+      Assert.That (driverConfigurationArgument.CommandTimeout, Is.EqualTo (configurationOverride.CommandTimeout));
```
MichaelKetting: test against explicit value. Reasoning: it would be possible for the webtesthelper to do the wrong thing (e.g. set the overrirde value instead of the new object) and the test would still be green.

Remotion/Web/Development.WebTesting/WebTestHelper.cs
```diff
@@ -190,12 +191,15 @@ public void OnSetUp ([NotNull] string testName)
...
     /// <param name="maximizeWindow">Specified whether the new browser session's window should be maximized.</param>
+    /// <param name="configurationOverride">Specifies additional options applied when creating the browser.</param>
     /// <returns>The new browser session.</returns>
-    public IBrowserSession CreateNewBrowserSession (bool maximizeWindow = true)
+    public IBrowserSession CreateNewBrowserSession (bool maximizeWindow = true, [CanBeNull] DriverConfigurationOverride configurationOverride = null)
     {
       using (new PerformanceTimer (s_log, string.Format ("Created new {0} browser session.", _browserConfiguration.BrowserName)))
       {
-        var browserResult = _browserConfiguration.BrowserFactory.CreateBrowser (_testInfrastructureConfiguration);
+        var mergedDriverConfiguration = MergeDriverConfiguration (_testInfrastructureConfiguration.DriverConfiguration, configurationOverride);
```
MichaelKetting: I think there was a discussion in the original PR about not exposing DriverCOnfiguration via TestInfrastructureCOnfiguration but separate. Please dig out the discussion so we can make the final decision on the design.

iam-mholle: > Should we keep DriverConfiguration on TestInfrastructureConfiguration or should split those types permanently? Or, from a reverse option, should we build the DriverConfiguration in WebTestHelper and also expose it on WebTestHelper.

Our decision would be to separate the `DriverConfiguration` from the `TestInfrastructureConfiguration` and expose it through the `WebTestHelper`. The `DriverConfiguration` seems equivalent to the `BrowserConfiguration` and `TestInfrastructureConfiguration`.

Remotion/Web/Development.WebTesting/Configuration/ITestInfrastructureConfiguration.cs
```diff
@@ -36,17 +36,6 @@ public interface ITestInfrastructureConfiguration
     /// </summary>
     string ScreenshotDirectory { [CanBeNull] get; }
 
-    /// <summary>
-    /// Specifies how long the Coypu engine should maximally search for a web element or try to interact with a web element before it fails.
-    /// </summary>
-    TimeSpan SearchTimeout { get; }
```
MichaelKetting: Document removed properties as two separate breaking changes.
When backporting, do not remove the properties from the interface. Instead, let them return the correct values from DriverConfiguration to avoid breaking existing code bases during a hotfix upgrade.

iam-mholle:
Added to master with 57c9dc2efb7894127db3504140d4274bc24a86c1.

PR #18, opened 2019-09-24
RM-7173: Opening 2 identical windows and resizing one throws exception in chrome

iam-mholle:


Remotion/Web/Development.WebTesting.IntegrationTests/MultiWindowTest.cs
```diff
@@ -27,6 +28,20 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests
...
   public class MultiWindowTest : IntegrationTest
   {
+    [Test]
+    public void TestMultiWindowResizeBrowserContent ()
+    {
+      var home = Start();
+      var newSize = new Size (600, 600);
+      var loadWindowFunctionInNewWindowButton = home.WebButtons().GetByID ("LoadWindowFunctionInNewWindow");
+      var window1 = loadWindowFunctionInNewWindowButton.Click().ExpectNewPopupWindow<WxePageObject> ("MyWindow");
+      var window2 = loadWindowFunctionInNewWindowButton.Click().ExpectNewPopupWindow<WxePageObject> ("MyWindow");
```
Stefan-Ilic: We could extract the window title into a constant variable to emphasize it's the same one twice

Remotion/Web/Development.WebTesting.IntegrationTests/MultiWindowTest.cs
```diff
@@ -27,6 +28,21 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests
...
+      const string windowTitle = "MyWindow";
+      var home = Start();
+      var newSize = new Size (600, 600);
+      var loadWindowFunctionInNewWindowButton = home.WebButtons().GetByID ("LoadWindowFunctionInNewWindow");
+      var window1 = loadWindowFunctionInNewWindowButton.Click().ExpectNewPopupWindow<WxePageObject> (windowTitle);
+      var window2 = loadWindowFunctionInNewWindowButton.Click().ExpectNewPopupWindow<WxePageObject> (windowTitle);
+
+      Helper.BrowserConfiguration.BrowserHelper.ResizeBrowserContentTo (window2.Context.Window, newSize);
+
+      Assert.That (Helper.BrowserConfiguration.BrowserHelper.GetBrowserContentBounds (window1.Context.Window).Size, Is.EqualTo (newSize));
```
MichaelKetting: Should we have a second test that tests that window 2 can also be selected and it not just happens to always select the first window?

iam-mholle:
I will need to investigate how coypu handles the window selection, there seems to be an issue. `window1.Context.Window` and `window2.Context.Window` point to different coypu browser objects, but interacting with them leads to interacting with the same chrome window.

PR #20, opened 2019-09-27
RM-7251: ClickAction should only call .Focus() before .Click() in InternetExplorer

iam-mholle:


iam-mholle:
Added to master with 0c2685bedf863a19a591106716e4334cbe6cc760.

PR #25, opened 2019-11-06
RM-7278: Disable chromedriver spec compliance

iam-mholle:
[RM-7278](https://re-motion.atlassian.net/browse/RM-7278)


iam-mholle:
Added to master with 2f5ad2d2a83a838686bd10f2ebf80426b3639a72.

PR #26, opened 2019-11-08
RM-7280: TestExecutionScreenshotRecorderTest.TearDown() does not check if the used path exists before accessing it

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7280


iam-mholle:
Added to master with 2673331b708ee85075c5d7c55fc093f4e9e8a057.

PR #31, opened 2019-11-20
RM-7295: ChromeUserDirectoryCleanUpStrategy should accept the userdirectory root as string instead of the whole IChromeConfiguration

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7295


iam-mholle:
Added to master with b75ac5350b31acca4246a4c5e044bad52959a888.

PR #32, opened 2019-11-25
RM-7172: Automated Accessibility Testing

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7172 

@MichaelKetting I believe the only things you haven't reviewed yet are the two new test projects and their integration into the solution.


Stefan-Ilic:
I just remembered:

- [ ] Update targetframework of new test projects to 4.6.1
- [ ] Check if newly thrown exceptions should be WebTestExceptions

MichaelKetting:
In *.Web.IntegrationTests:
Rename test fixture from "\BocXxx\Accessibility" to "\BocXxx\AccessibilityBoxXxxTest"

1. Test fixtures should end in "Test"
2. Try to prevent non-unique type names also in tests. Makes it easier to navigate the code.

MichaelKetting:
> 
> 
> I just remembered:
> 
>     * [ ]  Update targetframework of new test projects to 4.6.1
> 
>     * [ ]  Check if newly thrown exceptions should be WebTestExceptions

Plase rebase on top and fix those before the review.

Stefan-Ilic:
> > I just remembered:
> > ```
> > * [ ]  Update targetframework of new test projects to 4.6.1
> > 
> > * [ ]  Check if newly thrown exceptions should be WebTestExceptions
> > ```
> 
> Plase rebase on top and fix those before the review.

Done

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AccessibilityConformanceLevelConverter.cs
```diff
@@ -0,0 +1,82 @@
...
+                { "section508", AccessibilityConformanceLevel.Section508 },
+            };
+
+      s_enumToString = s_stringToEnum.ToDictionary (kvp => kvp.Value, kvp => kvp.Key);
+    }
+
+    /// <summary>
+    /// Converts a <see cref="string"/> to its <see cref="AccessibilityConformanceLevel"/> representation.
+    /// </summary>
+    public static AccessibilityConformanceLevel ConvertToEnum ([NotNull] string conformanceLevelAsString)
```
MichaelKetting: also annotate return values

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityAnalyzer.cs
```diff
@@ -0,0 +1,317 @@
...
+      return new AccessibilityAnalyzer (remoteWebDriver, remoteWebDriver, axeResultParser, configuration, sourceProvider, mapper, logger);
+    }
+
+    /// <summary>
+    /// Indicates whether IFrames are included.
+    /// </summary>
+    public bool IncludeIFrames { get; set; }
+
+    /// <summary>
+    /// Timeout of IFrames in milliseconds.
```
MichaelKetting: drop 'milliseconds'

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityAnalyzer.cs
```diff
@@ -0,0 +1,317 @@
...
+    public void IgnoreControlObject ([NotNull] ControlObject element)
+    {
+      ArgumentUtility.CheckNotNull ("element", element);
+
+      IgnoreCssSelector ($"#{element.GetHtmlID()}");
+    }
+
+    /// <summary>
+    /// Ignore control object with given css selector during the accessibility analysis.
+    /// <i>Does not work with the html element.</i>
```
MichaelKetting: no italics. either just second sentence or remarks.

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityAnalyzer.cs
```diff
@@ -0,0 +1,317 @@
...
+
+      return Mapper.Map (parsedResult);
+    }
+
+    private bool AxeIsInjected ()
+    {
+      return (bool) JsExecutor.ExecuteScript ("return (typeof axe !== 'undefined')");
+    }
+
+    // Injects Axe into every iframe manually because axe-core cannot do it itself
```
MichaelKetting: summary + remarks. don't use line comments outside of method bodies given how r# does not move them with the method during refactorings.

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityAnalyzer.cs
```diff
@@ -0,0 +1,317 @@
...
+
+    private void InjectAxeSource (string source)
+    {
+      JsExecutor.ExecuteScript (source);
+
+      if (Configuration.IncludeIFrames)
+        InjectIntoIFrames (source);
+    }
+
+    private string BuildAxeRunFunctionCall (string cssToInclude)
```
MichaelKetting: should we name the methods receiving a stringbuilder 'append' instead of 'build'?

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityAnalyzer.cs
```diff
@@ -0,0 +1,317 @@
...
+
+    private void BuildAxeRunFunctionCallContextExclude (StringBuilder stringBuilder, string cssToInclude)
+    {
+      if (ExcludedElements.Count != 0)
+      {
+        if (cssToInclude != null)
+          stringBuilder.Append (',');
+
+        stringBuilder.Append ("exclude: [");
+        stringBuilder.Append (string.Join (", ", ExcludedElements.Select (s => $"['{s}']")));
```
MichaelKetting: use aggregate with string builder.

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityAnalyzer.cs
```diff
@@ -0,0 +1,317 @@
...
+          InjectAxeSource (source);
+        }
+      }
+
+      var axeRunFunctionCall = BuildAxeRunFunctionCall (cssSelector);
+
+      string result;
+      using (new PerformanceTimer (Logger, "Accessibility analysis has been performed."))
+      { 
+        result = (string) JsExecutor.ExecuteAsyncScript (axeRunFunctionCall) ?? GetRawAxeResultForInternetExplorer (axeRunFunctionCall);
```
MichaelKetting: am not sure I want to use null coallese here. two non-primitive statements chained and such. what does the rest of the code say about this style?

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityConfiguration.cs
```diff
@@ -0,0 +1,90 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+using System.Text;
+using Remotion.Web.Development.WebTesting.Accessibility.Implementation;
+
+namespace Remotion.Web.Development.WebTesting.Accessibility
+{
+  /// <inheritdoc cref="IAccessibilityConfiguration" />
```
MichaelKetting: Default impl...

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityConfiguration.cs
```diff
@@ -0,0 +1,90 @@
...
+  {
+    public AccessibilityConformanceLevel ConformanceLevel { get; }
+    public bool IncludeIFrames { get; }
+    public TimeSpan IFrameTimeout { get; }
+    public bool ScrollToInitialPosition { get; }
+    public bool AbsolutePaths { get; }
+    public bool XPath { get; }
+
+    public AccessibilityConfiguration (
+        AccessibilityConformanceLevel? conformanceLevel = null,
```
MichaelKetting: why use nullable instead of the value when we can use value type?

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityConfiguration.cs
```diff
@@ -0,0 +1,90 @@
...
+        AppendWcag21A (sb);
+        AppendWcag2DoubleA (sb);
+        AppendExperimental (sb);
+      }
+
+      sb.Append ("]}");
+
+      return sb.ToString();
+
+      void AppendWcag2A (StringBuilder stringBuilder) => stringBuilder.Append (",\"wcag2a\"");
```
MichaelKetting: good choice with the local methods

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityResult.cs
```diff
@@ -0,0 +1,131 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using JetBrains.Annotations;
+using Remotion.Utilities;
+
+namespace Remotion.Web.Development.WebTesting.Accessibility
+{
+  /// <summary>
+  /// Represents the Result of the Accessibility Analysis in a readable format.
```
MichaelKetting: casing

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityResult.cs
```diff
@@ -0,0 +1,131 @@
...
+    /// </summary>
+    public string OrientationType { get; }
+
+    /// <summary>
+    /// User agent that was used.
+    /// </summary>
+    public string UserAgent { get; }
+
+    /// <summary>
+    /// Whether <see cref="IAccessibilityConfiguration.IncludeIFrames">IFrames were included</see> during the test.
```
MichaelKetting: use of see is non-typical. reword.

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityResultFilter.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// <summary>
+    /// A list of rule IDs that should be ignored.
+    /// </summary>
+    public List<AccessibilityRuleID> IgnoreRuleID { get; } = new List<AccessibilityRuleID>();
+
+    /// <summary>
+    /// The impact to filter by.
+    /// </summary>
+    public AccessibilityTestImpact? IncludeImpact { get; set; }
+  }
```
MichaelKetting: ctor

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityRequirement.cs
```diff
@@ -0,0 +1,45 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+using Remotion.Utilities;
+
+namespace Remotion.Web.Development.WebTesting.Accessibility
+{
+  /// <summary>
+  /// Contains information about the accessibility check.
```
MichaelKetting: should this be 'requirement'?

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityTestImpact.cs
```diff
@@ -0,0 +1,32 @@
...
+// 
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+
+namespace Remotion.Web.Development.WebTesting.Accessibility
+{
+  /// <summary>
+  /// The Impact of an Accessibility Check, Rule or Node.
```
MichaelKetting: casing. check all xml docs

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityTestSuccessCriteria.cs
```diff
@@ -0,0 +1,715 @@
...
+
+namespace Remotion.Web.Development.WebTesting.Accessibility
+{
+  /// <summary>
+  /// Accessibility standards.
+  /// </summary>
+  public enum AccessibilityTestSuccessCriteria
+  {
+    /// <summary>
+    /// <b>Non-text Content</b>
```
MichaelKetting: split into summary and remarks. can we use search and replace here? maybe a bit of regex to help with the start-tag?

Remotion/Web/Development.WebTesting/Accessibility/ControlObjectExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+using JetBrains.Annotations;
+using Remotion.Utilities;
+
+namespace Remotion.Web.Development.WebTesting.Accessibility
+{
+  /// <summary>
+  /// Accessibility extension class for ControlObject.
```
MichaelKetting: see cref

Remotion/Web/Development.WebTesting/Accessibility/IAccessibilityConfiguration.cs
```diff
@@ -0,0 +1,51 @@
...
+// 
+using System;
+using JetBrains.Annotations;
+
+namespace Remotion.Web.Development.WebTesting.Accessibility
+{
+  /// <summary>
+  /// Accessibility Analysis settings.
+  /// </summary>
+  public interface IAccessibilityConfiguration
```
MichaelKetting: /// <seealso

Remotion/Web/Development.WebTesting/Accessibility/PageObjectExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+using System;
+using JetBrains.Annotations;
+using Remotion.Utilities;
+
+namespace Remotion.Web.Development.WebTesting.Accessibility
+{
+  /// <summary>
+  /// Accessibility extension class for PageObject.
+  /// </summary>
+  public static class PageObjectExtensions
```
MichaelKetting: AccessibilityPageObjectExtensions

Remotion/Web/Development.WebTesting/Accessibility/IAxeSourceProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+using JetBrains.Annotations;
+
+namespace Remotion.Web.Development.WebTesting.Accessibility
+{
+  /// <summary>
+  /// Provides the axe js API.
```
MichaelKetting: Defines an API for retrieving the <c>axe.js</c> file contents to be used for accessibility analysis.

Remotion/Web/Development.WebTesting.IntegrationTests/Accessibility/AccessibilityAnalyzerTest.cs
```diff
@@ -0,0 +1,114 @@
...
+
+      var result = home.Analyze (analyzer);
+      var violation = result.Violations.Single();
+
+      Assert.That (violation.Rule.ID, Is.EqualTo (AccessibilityRuleID.Label));
+    }
+
+
+    [Test]
+    public void Analyze_ExcludeRule ()
```
MichaelKetting: Hmm...I think I want to have a non-empty result at the end of the exclude tests. otherwise, we don't know if the exclude removed just the specified rules or all rules.

Remotion/Web/Development.WebTesting.IntegrationTests/Accessibility/AccessibilityIFrameSupportTest.cs
```diff
@@ -0,0 +1,39 @@
...
+  public class AccessibilityIFrameSupportTest : IntegrationTest
+  {
+    [Test]
+    public void TestInjectSourceToIFrames ()
+    {
+      var home = Start<WxePageObject> ("Accessibility/IFrame.html");
+      var analyzer = Helper.CreateAccessibilityAnalyzer();
+
+      var accessibilityResult = home.Analyze (analyzer);
+      var violations = accessibilityResult.Violations.IgnoreByCssSelector ("html");
```
MichaelKetting: am confused. also, wasn't there a not that ignore doesn't work for 'html'. am not 100% sure which exclude/filter that was exactly, but there was a note/guard somewhere.

Remotion/Web/Development.WebTesting.IntegrationTests/Accessibility/AccessibilityResultTest.cs
```diff
@@ -0,0 +1,109 @@
...
+  public class AccessibilityResultTest : IntegrationTest
+  {
+    [Test]
+    public void IgnoreByCssSelector ()
+    {
+      Start();
+      var analyzer = Helper.CreateAccessibilityAnalyzer();
+      var result = analyzer.Analyze();
+
+      var violations = result.Violations.IgnoreByCssSelector ("html");
```
MichaelKetting: see note for iframe support

Remotion/Web/Development.WebTesting.IntegrationTests/Accessibility/AccessibilityAnalyzerTest.cs
```diff
@@ -0,0 +1,114 @@
...
+      var analyzer = Helper.CreateAccessibilityAnalyzer();
+
+      analyzer.IgnoreRule (AccessibilityRuleID.Label);
+      var result = analyzer.Analyze();
+
+      Assert.That (result.Violations, Is.Empty);
+    }
+
+    [Test]
+    public void Analyze_ExcludeMultipleRules ()
```
MichaelKetting: should those tests use "Ignore" instead of "Exclude" in the name. check other fixtures, too.

Remotion/Web/Development.WebTesting.IntegrationTests/Accessibility/AccessibilityViolationAnalysisTest.cs
```diff
@@ -0,0 +1,271 @@
...
+      Assert.That (violation.Rule.SuccessCriteria, Has.Length.EqualTo (4));
+      Assert.That (violation.Rule.SuccessCriteria, Contains.Item (AccessibilityTestSuccessCriteria.Cat_forms));
+      Assert.That (violation.Rule.SuccessCriteria, Contains.Item (AccessibilityTestSuccessCriteria.Wcag_3_3_2));
+      Assert.That (violation.Rule.SuccessCriteria, Contains.Item (AccessibilityTestSuccessCriteria.Wcag_1_3_1));
+      Assert.That (violation.Rule.SuccessCriteria, Contains.Item (AccessibilityTestSuccessCriteria.Section508_22_n));
+      Assert.That (violation.TargetPath.Single().XPath, Is.EqualTo ("/input[@id='input']"));
+    }
+
+    [Test]
+    public void AccessibilityImageWithoutAltTest ()
```
MichaelKetting: "test" at the end?

Remotion/Web/Development.WebTesting.IntegrationTests/TabbedMultiViewControlObjectTest.cs
```diff
@@ -116,26 +124,43 @@ public void TestSwitchTo ()
       tabbedMultiView = home.TabbedMultiViews().Single();
       Assert.That (tabbedMultiView.Scope.Text, Is.StringContaining ("Content2"));
       Assert.That (tabbedMultiView.Scope.Text, Is.Not.StringContaining ("Content1"));
+      Assert.That (tabbedMultiView.Scope.Text, Is.Not.StringContaining ("Content3"));
```
MichaelKetting: why do we need the tab3 tests here?

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityConfiguration.cs
```diff
@@ -0,0 +1,90 @@
...
+{
+  /// <inheritdoc cref="IAccessibilityConfiguration" />
+  public class AccessibilityConfiguration : IAccessibilityConfiguration
+  {
+    public AccessibilityConformanceLevel ConformanceLevel { get; }
+    public bool IncludeIFrames { get; }
+    public TimeSpan IFrameTimeout { get; }
+    public bool ScrollToInitialPosition { get; }
+    public bool AbsolutePaths { get; }
+    public bool XPath { get; }
```
MichaelKetting: Use "Enable..." or "Include..." etc for flags.

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityAnalyzerTest.cs
```diff
@@ -0,0 +1,273 @@
...
+    public void CreateAnalyzer_ChangesDefaultConfiguration ()
+    {
+      _configStub
+          .Stub (_ => _.IFrameTimeout)
+          .Return (TimeSpan.FromSeconds (5));
+      _configStub
+          .Stub (_ => _.IncludeIFrames)
+          .Return (true);
+
+      var analyzer = new TestableAccessibilityAnalyzer (
```
MichaelKetting: I think we should have a factory method. this is a lot of duplication maintenance.

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityAnalyzerTest.cs
```diff
@@ -0,0 +1,273 @@
...
+          _sourceProviderStub,
+          _resultMapperStub,
+          _loggerStub);
+
+      Assert.That (analyzer.IncludeIFrames, Is.True);
+      Assert.That (analyzer.IFrameTimeout, Is.EqualTo (TimeSpan.FromSeconds (5)));
+    }
+
+    [Test]
+    public void CreateAnalyzer_ChangesDefaultConfiguration ()
```
MichaelKetting: Initialize

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityAnalyzerTest.cs
```diff
@@ -0,0 +1,273 @@
...
+          "version",
+          0,
+          0,
+          0,
+          "orientationType",
+          "userAgent",
+          true,
+          AccessibilityConformanceLevel.Wcag20_ConformanceLevelDoubleA,
+          new List<AccessibilityRuleResult>());
+      _jsExecutorStub
```
MichaelKetting: can use single line if there's enough line width available. linewrap only if there's long statements or expected to become long statements.

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityCheckIDConverterTest.cs
```diff
@@ -0,0 +1,35 @@
...
+using Remotion.Web.Development.WebTesting.Accessibility;
+using Remotion.Web.Development.WebTesting.Accessibility.Implementation;
+
+namespace Remotion.Web.Development.WebTesting.UnitTests.Accessibility
+{
+  [TestFixture]
+  public class AccessibilityCheckIDConverterTest
+  {
+    [Test]
+    public void GetEnum_UnknownKey_Unknown ()
```
MichaelKetting: known key, too.

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityConfigurationTest.cs
```diff
@@ -0,0 +1,96 @@
...
+          "iframes: false, xpath: true, frameWaitTime: 13000, absolutePaths: true, restoreScroll: true, runOnly: {type:\"tag\",values:[\"wcag2a\"]}";
+
+      var configuration = new AccessibilityConfiguration (
+          AccessibilityConformanceLevel.Wcag20_ConformanceLevelA,
+          false,
+          TimeSpan.FromSeconds (13),
+          true,
+          true);
+
+      var result = configuration.GetJson();
```
MichaelKetting: why does the json not have braces?

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityConfigurationTest.cs
```diff
@@ -0,0 +1,96 @@
...
+    [Test]
+    public void CreateConfiguration_NegativeIFrameTimeout_ThrowsArgumentOutOfRangeException ()
+    {
+      Assert.That (
+          () => new AccessibilityConfiguration (iframeTimeout: TimeSpan.FromSeconds (-5)),
+          Throws.TypeOf<ArgumentOutOfRangeException>());
+    }
+
+    [Test]
+    public void CreateConfiguration_NoCtorParameters_DefaultValues ()
```
MichaelKetting: "Initialize"

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityConformanceLevelConverterTest.cs
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Remotion.Web.Development.WebTesting.UnitTests.Accessibility
+{
+  [TestFixture]
+  public class AccessibilityConformanceLevelConverterTest
+  {
+    [Test]
+    public void ConvertToEnum_UnknownKey_Unknown ()
+    {
+      var conformanceLevel = AccessibilityConformanceLevelConverter.ConvertToEnum ("SomeUnknownKey");
```
MichaelKetting: + good value test

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AxeToolOptionsRunOnly.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Remotion.Web.Development.WebTesting.Accessibility.Implementation
+{
+  /// <summary>
+  /// Used by the JSON deserializer.
+  /// </summary>
+  [DataContract]
+  public class AxeToolOptionsRunOnly
+  {
+    /// <summary>Either tags or rules.</summary>
```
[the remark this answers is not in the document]
iam-mholle: Sorry, I do not understand. Could you elaborate?

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityResultMapperTest.cs
```diff
@@ -0,0 +1,225 @@
...
+using Remotion.Web.Development.WebTesting.Accessibility;
+using Remotion.Web.Development.WebTesting.Accessibility.Implementation;
+
+namespace Remotion.Web.Development.WebTesting.UnitTests.Accessibility
+{
+  [TestFixture]
+  public class AccessibilityResultMapperTest
+  {
+    [Test]
+    public void ConvertToFinalResult_ParseSuccessCriteriaTags ()
```
MichaelKetting: "Map"

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AxeToolOptionsRunOnly.cs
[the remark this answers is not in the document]
iam-mholle: Nevermind, got it

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityResultViolationExtensionTest.cs
```diff
@@ -0,0 +1,225 @@
...
+            AccessibilityTestImpact.Moderate,
+            new[] { AccessibilityTestSuccessCriteria.Wcag_1_3_6 }),
+        new[] { new AccessibilityTestTarget ("xPath", "CSS") },
+        "Html",
+        new[] { new AccessibilityRequirement (AccessibilityCheckID.MetaViewport, "Message", AccessibilityTestImpact.Moderate) },
+        new AccessibilityRequirement[0],
+        new AccessibilityRequirement[0]);
+
+    [Test]
+    public void FilterByCssSelectorAndImpact ()
```
MichaelKetting: Filter_By...

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityResultViolationExtensionTest.cs
```diff
@@ -0,0 +1,225 @@
...
+using System.Linq;
+using NUnit.Framework;
+using Remotion.Web.Development.WebTesting.Accessibility;
+
+namespace Remotion.Web.Development.WebTesting.UnitTests.Accessibility
+{
+  [TestFixture]
+  public class AccessibilityResultViolationExtensionTest
+  {
+    private readonly AccessibilityRuleResult _defaultViolation = new AccessibilityRuleResult (
```
MichaelKetting: Setup

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityResultViolationExtensionTest.cs
```diff
@@ -0,0 +1,225 @@
...
+    {
+      var filter = new AccessibilityResultFilter
+                   {
+                       IgnoreCssSelectors = { "test" },
+                       IncludeImpact = AccessibilityTestImpact.Moderate
+                   };
+      var violation = CreateViolation (impact: AccessibilityTestImpact.Moderate, targets: new[] { CreateTargetWithXPathAndCss ("test") });
+      var accessibilityResult = CreateAccessibilityResult (violation, _defaultViolation);
+
+      var filteredViolations = accessibilityResult.Violations.Filter (filter);
```
MichaelKetting: Do we actually need a accessibilityresult or should we just have a collection of the correct type?

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityResultViolationExtensionTest.cs
```diff
@@ -0,0 +1,225 @@
...
+      var violation = CreateViolation (checkIDs: new[] { AccessibilityCheckID.ThHasDataCells, AccessibilityCheckID.DuplicateId });
+      var accessibilityResult = CreateAccessibilityResult (violation, _defaultViolation);
+
+      var filteredViolations = accessibilityResult.Violations.IgnoreByCheckID (AccessibilityCheckID.ThHasDataCells, AccessibilityCheckID.DuplicateId);
+
+      Assert.That (filteredViolations, Is.EquivalentTo (new[] { _defaultViolation }));
+    }
+
+    [Test]
+    public void IgnoreByRuleIDAndCssSelector ()
```
MichaelKetting: name

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityResultViolationExtensionTest.cs
```diff
@@ -0,0 +1,225 @@
...
+      var violation = CreateViolation (targets: new[] { CreateTargetWithXPathAndCss ("test1") });
+      var accessibilityResult = CreateAccessibilityResult (violation, _defaultViolation);
+
+      var filteredViolations = accessibilityResult.Violations.IgnoreByCssSelector ("test1");
+
+      Assert.That (filteredViolations, Is.EquivalentTo (new[] { _defaultViolation }));
+    }
+
+    [Test]
+    public void IgnoreByMultipleCheckIDs ()
```
MichaelKetting: name

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityResultViolationExtensionTest.cs
```diff
@@ -0,0 +1,225 @@
...
+
+      Assert.That (filteredViolations, Is.EquivalentTo (new[] { _defaultViolation }));
+    }
+
+    [Test]
+    public void IgnoreByRuleIDAndMultipleCssSelectors ()
+    {
+      var violation = CreateViolation (AccessibilityRuleID.AreaAlt, targets: new[] { CreateTargetWithXPathAndCss ("testFrame"), CreateTargetWithXPathAndCss ("test") });
+      var accessibilityResult = CreateAccessibilityResult (violation, _defaultViolation);
+
```
MichaelKetting: name. check remaining mthods, too

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityRuleIDConverterTest.cs
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Remotion.Web.Development.WebTesting.UnitTests.Accessibility
+{
+  [TestFixture]
+  public class AccessibilityRuleIDConverterTest
+  {
+    [Test]
+    public void GetEnum_UnknownKey_Unknown ()
+    {
+      var ruleID = AccessibilityRuleIDConverter.ConvertToEnum ("SomeUnknownKey");
```
MichaelKetting: happy path value tes  ,too

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AxeResultParserTest.cs
```diff
@@ -0,0 +1,255 @@
...
+    [TestCase ("{[[}")]
+    public void ParseResults_InvalidFormat_ThrowsException (string jsonResponse)
+    {
+      var parser = new AxeResultParser();
+
+      Assert.That (() => parser.Parse (jsonResponse), Throws.InstanceOf<SerializationException>());
+    }
+
+    [Test]
+    public void ParseResults_ParseAccessibilityTestResultViolations ()
```
MichaelKetting: name: Parse_WithTestResultContainsViolations.

check remaining names

Remotion/ObjectBinding/Web.IntegrationTests/BocTreeView/AccessibilityBocTreeViewTest.cs
```diff
@@ -0,0 +1,131 @@
...
+      var bocTreeView = home.TreeViews().GetByLocalID ("NoPropertyIdentifier");
+      var analyzer = Helper.CreateAccessibilityAnalyzer();
+
+      var result = bocTreeView.Analyze (analyzer);
+
+      Assert.That (result.Violations, Is.Empty);
+    }
+
+    [Test]
+    public void ContextMenuDelayed ()
```
MichaelKetting: I think we may have a few too many testcases here

Remotion/ObjectBinding/Web.IntegrationTests/ObjectBinding.Web.IntegrationTests.csproj
```diff
@@ -0,0 +1,95 @@
...
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="nunit.framework, Version=2.6.2.12296, Culture=neutral, PublicKeyToken=96d09a1eb7f44a77, processorArchitecture=MSIL">
+      <HintPath>..\..\..\packages\NUnit.2.6.2\lib\nunit.framework.dll</HintPath>
+    </Reference>
+    <Reference Include="System" />
+    <Reference Include="System.Core" />
+    <Reference Include="System.Drawing" />
+    <Reference Include="System.Xml.Linq" />
```
MichaelKetting: do we need all those references?

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AccessibilityConformanceLevelConverter.cs
iam-mholle: What do you mean? `AccessibilityConformanceLevel` is an `enum`.

Remotion/ObjectBinding/Web.IntegrationTests/BocTreeView/AccessibilityBocTreeViewTest.cs
Stefan-Ilic: I believe each of them represents some state of `BocList`, no? Which one would you remove? Couldn't we also leave them if they don't increase maintenance effort?

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityConfigurationTest.cs
Stefan-Ilic: It doesn't return a JSON object, rather the state of the object as "json body content". It's the format the `AccessibilityAnalyzer` needs for injecting. Do we

- [ ] Add braces here and remove them in `Analyzer` (where we need them) ?)
- [ ] Rename the method to `GetJsonContent`
- [ ] None of the above

Remotion/Web/Development.WebTesting.IntegrationTests/TabbedMultiViewControlObjectTest.cs
Stefan-Ilic: It looks like a tab with an icon was added to test its accessibility. I'm guessing we added this to this test to keep it "complete"

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityCheckID.cs
```diff
@@ -0,0 +1,107 @@
...
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+
+namespace Remotion.Web.Development.WebTesting.Accessibility
+{
+  /// <summary>
+  /// Accessibility Check IDs.
+  /// </summary>
+  public enum AccessibilityCheckID
```
iam-mholle: Should this be `AccessibilityRequirementId`?

Remotion/ObjectBinding/Web.IntegrationTests/ObjectBinding.Web.IntegrationTests.csproj
Stefan-Ilic: I'm guessing something that we got rid of needed them, I removed all except one and it all seems to work fine

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityResultViolationExtensionTest.cs
Stefan-Ilic: What do you mean exactly? Do you want to

- [ ] Get rid of the CreateAccessibilityResult factory method in these tests?
- [ ] Remove the type `AccessibilityResult` overall?
- [ ] None of the above

MichaelKetting: 1. Rename CreateAccessiblityResult to CreateViolations / alternatively, just create an array with the relevant entries if this remains legible. 
2. change var accessiblityResult to hold the collection of Violation objects.
3. Apply Filter() to this variable.
4. Remove no longer needed code from Test, do not change production code.

Remotion/Web/Development.WebTesting/Accessibility/AccessibilityCheckID.cs
MichaelKetting: Upper-case ID. Check vs Requirement. don't know. need input.

Remotion/Web/Development.WebTesting.IntegrationTests/TabbedMultiViewControlObjectTest.cs
MichaelKetting: kill it in this test. It feels redundant to do same thing 3 times.

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityConfigurationTest.cs
MichaelKetting: Okay, so is our code the one that doesn't want braces or is that actually what happens inside axe.js? if it's our code, can we change that?

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AccessibilityConformanceLevelConverter.cs
MichaelKetting: And how am I supposed to know that from reading black and white text without any intellisense? :D

Remotion/ObjectBinding/Web.IntegrationTests/BocTreeView/AccessibilityBocTreeViewTest.cs
MichaelKetting: too many redundant cases increase the test run time. test cases that use the same markup and only differ in some behavior can be considered redundant. e.g. testing two input controls, one with a server side validator and one without will be redundant until I also test the validation result. Have to check for those. probably will want to make a suggestion and then ask me for a complete list :)

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AxeSourceProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+  /// <inheritdoc cref="IAxeSourceProvider" />
+  public class AxeSourceProvider : IAxeSourceProvider
+  {
+    private const string c_resourceName = "Remotion.Web.Development.WebTesting.Accessibility.axe.min.js";
+
+    public string GetSource ()
+    {
+      var assembly = Assembly.GetExecutingAssembly();
+
+      using (var stream = assembly.GetManifestResourceStream (c_resourceName))
```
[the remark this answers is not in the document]
MichaelKetting: Todo

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AxeRuleNodesAnyRelatedNodes.cs
```diff
@@ -0,0 +1,44 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+// 
+using System;
+using System.Runtime.Serialization;
+
+namespace Remotion.Web.Development.WebTesting.Accessibility.Implementation
+{
+  /// <summary>
+  /// Represents the 'relatedNodes' JSON object in an AXE analysis result.
```
MichaelKetting:
``` 
<c> instead of '

File: Remotion/Web/Development.WebTesting/Accessibility/IAccessibilityConfiguration.cs [hunk 50]
```diff
@@ -0,0 +1,52 @@
...
+    bool EnableAbsolutePaths { get; }
+
+    /// <summary>Whether the axe result should contain XPaths.</summary>
+    bool EnableXPath { get; }
+
+    /// <summary>
+    /// Creates Json with specified options.
+    /// </summary>
+    [NotNull]
+    string GetJson ();
```
MichaelKetting: Am not sure if GetJson and config-options should be on the same interface. Possibly, also not the same implemetnation.

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityRuleIDConverterTest.cs
```diff
@@ -0,0 +1,51 @@
...
+
+      Assert.That (ruleID, Is.EqualTo (AccessibilityRuleID.Unknown));
+    }
+
+    [Test]
+    public void GetEnum_KnownKey_KeyAsEnum ()
```
MichaelKetting: GetEnum_WithKnownKey_ReturnsKeyAsEnum.

Other tests, too, if applicable.

iam-mholle:
Still WIP.

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AccessibilityResultMapper.cs
```diff
@@ -0,0 +1,134 @@
...
+    private IEnumerable<AccessibilityTestSuccessCriteria> GetSuccessCriteria (string[] axeRuleTags)
+    {
+      return axeRuleTags
+          .Where (t => !AccessibilityConformanceLevelConverter.IsValid (t))
+          .Select (ParseAccessibilityTestSuccessCriteria);
+    }
+
+    private AccessibilityTestSuccessCriteria ParseAccessibilityTestSuccessCriteria (string rawTag)
+    {
+      var formattedTag = Regex.Replace (rawTag, @"wcag(\d)(\d)(\d)", "wcag_$1_$2_$3")
```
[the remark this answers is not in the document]
iam-mholle: > should the regex use begin and end specifiers?

Will do.

> and qualify that the last digit-group can be 1 or 2 digis?

Axe does not define any tags matching `wcag\d{4,}`.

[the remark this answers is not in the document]
MichaelKetting: > Axe does not define any tags matching wcag\d{4,}.
Yeah, but there's 2 digits in the third nesting level possible and we don't want to have silent fails.

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AccessibilityResultMapper.cs
```diff
@@ -0,0 +1,134 @@
...
+          .Replace ('.', '_').Replace ('-', '_');
```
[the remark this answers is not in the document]
iam-mholle: No, there are still conversions needed for e.g. `"section508.22.n"` -> `AccessibilityTestSuccessCriteria.Section508_22_n`.

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AccessibilityResultMapper.cs
```diff
@@ -0,0 +1,134 @@
...
+          "yyyy-MM-ddTHH:mm:ss.fffZ",
+          CultureInfo.CreateSpecificCulture ("en-US"),
+          DateTimeStyles.AdjustToUniversal);
+    }
+
+    private AccessibilityConformanceLevel ParseConformanceLevel (string[] tags)
+    {
+      var conformanceLevels = tags.Where (AccessibilityConformanceLevelConverter.IsValid);
+
+      if (tags.Contains ("section808"))
```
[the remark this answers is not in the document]
iam-mholle: It must have been a typo, either way, this if statement has no effect.

What should be the behavior when both (any) WCAG & Section508 are requested conformance levels? Should this be allowed?
What should be the (singular) conformance level in the output? Or should it be changed to an `IReadOnlyList`?

Remotion/Web/Development.WebTesting.IntegrationTests/Accessibility/AccessibilityAnalyzerTest.cs
iam-mholle: Would this also be needed for multiple excludes/ignores?

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AccessibilityResultMapper.cs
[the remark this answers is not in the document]
MichaelKetting: hmm...can we do with with regex, too. that way it would be easier to understand why there's two of them.

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AccessibilityResultMapper.cs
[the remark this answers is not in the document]
MichaelKetting: I think we decided to not support enabling section508. anyway, right now, we only support doing one conformance level at a time since that's what we actually need.

Remotion/Web/Development.WebTesting.IntegrationTests/Accessibility/AccessibilityAnalyzerTest.cs
MichaelKetting: yeah, think so.

Remotion/Web/Development.WebTesting.UnitTests/Accessibility/AccessibilityConfigurationTest.cs
Stefan-Ilic: Okay, so this is where it's used

```
    private void AppendAxeRunFunctionCallOptions (StringBuilder stringBuilder)
    {
      stringBuilder.Append ('{');
      stringBuilder.Append (Configuration.GetJson());

      if (ExcludedRules.Any())
      {
        stringBuilder.Append (",rules:{");
        stringBuilder.Append (ExcludedRules.Select (x => '"' + x + "\": {enabled: false}").Aggregate ((i, j) => i + ",\n" + j));
        stringBuilder.Append ('}');
      }
    }
```

I suppose it's just our code. If we changed the StringBuilder magic inside `AccessibilityAnalyzer` a bit the config could return a 'complete' json object if so desired

MichaelKetting: Please do. Makes reasoning about the string that much easier.

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AccessibilityResultMapper.cs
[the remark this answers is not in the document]
iam-mholle: > I think we decided to not support enabling section508

I added dc10c549effed25febf7b77ea58d9d9b77feb6a5.

Remotion/Web/Development.WebTesting/Accessibility/IAccessibilityConfiguration.cs
Stefan-Ilic: Perhaps `IAccessibilityConfiguration` could implement `ISerializable` and we'd just use a serializer to get the JSON representation?

Remotion/Web/Development.WebTesting.IntegrationTests/Accessibility/AccessibilityIFrameSupportTest.cs
iam-mholle: Ignore does work for `html`.

It was necessary here because the page containing the iframes had errors which had to be ignored. As the errors have been fixed, `.IgnoreByCssSelector ("html")` is obsolete now.

Remotion/Web/Development.WebTesting/Accessibility/IAccessibilityConfiguration.cs
MichaelKetting: No, ISerializable is not meant to this. If, then you'd use a DataContract and the .NET JSON Serializer. Still would probably prefer to use separate Types, one the configuration and one the DTO to be json-serialized.

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AccessibilityResultMapper.cs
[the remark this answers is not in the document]
MichaelKetting: yes, separte enums could make sense. depends a bit on just how well we can name things, how much overhead this incurs in terms of code, etc.

Remotion/Web/Development.WebTesting.IntegrationTests/Accessibility/AccessibilityIFrameSupportTest.cs
MichaelKetting: okay, so we can remove it?

Remotion/Web/Development.WebTesting/Accessibility/IAccessibilityConfiguration.cs
Stefan-Ilic: Are you looking for some more specific `IJsonifiableAccessibilityConfiguration` or a more generic `IJsonSerializable` for re-motion?

MichaelKetting: No to either. option 1) we just move the json-building to a different type, either an IAccessibleConfigurationJsonSerializer that acceepts the IAccessiblityConfiguration and returns a string, or we do it as a private code at the callsite of GetJson. Option 2) we make a DTO type, populate it and json-serialize, also at the callside of GetJson. It really depends on how and where we want to assert the json-string.

Remotion/Web/Development.WebTesting.IntegrationTests/Accessibility/AccessibilityIFrameSupportTest.cs
iam-mholle: Just to clarify my last comment:

`AccessibilityAnalyser.IgnoreCssSelector(...)` does not work with the `html` element (axe does not support it), this is also specified in its docs. The method that was used here was a filter method we're providing:

```csharp
public static IEnumerable<AccessibilityRuleResult> IgnoreByCssSelector (
    [NotNull] this IEnumerable<AccessibilityRuleResult> violations,
    [NotNull] params string[] cssSelector)
```

Remotion/Web/Development.WebTesting/Accessibility/IAccessibilityConfiguration.cs
Stefan-Ilic: In that case, to keep it simple I would probably just do it as private code at callside since the unit tests only cover the trivial scenarios and the interesting edge cases currently all need an integration test with real axe anyway

MichaelKetting: Okay...?

As for what's needed: it's a matter of syntax. Even with out AXE / integration test we can verify that the complicated scenarios generate syntactically correct json. this alone might be worth doing dedicated tests for. Otherwise, we depend on not making mistakes and if there's mistakes, the AXE / JS-engine giving us at least error detection if not helpful details.

iam-mholle:
@MichaelKetting Which conformance level should be used with re-motion controls?

MichaelKetting:
> 
> 
> @MichaelKetting Which conformance level should be used with re-motion control tests?

@iam-mholle Hmm...good point. WCAG 2.0 AA is mandatory. If we get good results with WCAG 2.1 AA, we should use that from the get-go.

Remotion/ObjectBinding/Web.IntegrationTests/BocDateTimeValue/AccessibilityBocDateTimeValueTest.cs
```diff
@@ -0,0 +1,84 @@
...
+      var bocDateTimeValue = home.DateTimeValues().GetByLocalID ("DateOfBirthField_Disabled");
+      var analyzer = Helper.CreateAccessibilityAnalyzer();
+
+      var result = bocDateTimeValue.Analyze (analyzer);
+
+      Assert.That (result.Violations, Is.Empty);
+    }
+
+    [Test]
+    public void DateOnly_ExpandedDatePicker ()
```
MichaelKetting: The DateOnly and ReadOnlyDateOnly tests we need in case rendering less introduces invalid markup (e.g. a missing reference to the not-rendered time field).

Remotion/ObjectBinding/Web.IntegrationTests/BocEnumValue/AccessibilityBocEnumValueTest.cs
```diff
@@ -0,0 +1,291 @@
...
+using NUnit.Framework;
+using Remotion.ObjectBinding.Web.Development.WebTesting.IntegrationTests;
+using Remotion.Web.Development.WebTesting.Accessibility;
+using Remotion.Web.Development.WebTesting.ExecutionEngine.PageObjects;
+using Remotion.Web.Development.WebTesting.FluentControlSelection;
+
+namespace Remotion.ObjectBinding.Web.IntegrationTests.BocEnumValue
+{
+  [TestFixture]
+  public class AccessibilityBocEnumValueTest : IntegrationTest
```
MichaelKetting: Validate-Test got dropped. Why. Actually, do we have Accessibility Tests for "a validation error occured" with the other control tests? if not, we should add them to ensure the markup that includes the business-logic-validation errors is valid.

Remotion/ObjectBinding/Web.IntegrationTests/BocList/AccessibilityBocListTest.cs
```diff
@@ -0,0 +1,121 @@
...
+using Remotion.Web.Development.WebTesting.ExecutionEngine.PageObjects;
+using Remotion.Web.Development.WebTesting.FluentControlSelection;
+
+namespace Remotion.ObjectBinding.Web.IntegrationTests.BocList
+{
+  [TestFixture]
+  public class AccessibilityBocListTest : IntegrationTest
+  {
+    [Test]
+    public void Normal ()
```
MichaelKetting: Empty-test is needed in case the the markup is different. Does the Empty test use the placeholder text or just doesn't show data-rows? We should have tests for both versions.

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AccessibilityConfigurationJsonSerializer.cs
```diff
@@ -0,0 +1,48 @@
...
+        AppendWcag2A (sb);
+
+      //axe version 3.2.2 requires the experimental tag for WCAG 2.1 A and WCAG 2.1 AA
+      if (configuration.ConformanceLevel == AccessibilityConformanceLevel.Wcag21_ConformanceLevelA)
+      {
+        AppendWcag2A (sb);
+        AppendExperimental (sb);
+      }
+
+      if (configuration.ConformanceLevel == AccessibilityConformanceLevel.Wcag21_ConformanceLevelDoubleA)
```
MichaelKetting: Test missing?

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AccessibilityConfigurationJsonSerializer.cs
```diff
@@ -0,0 +1,48 @@
...
+      sb.Append ($", absolutePaths: {configuration.EnableAbsolutePaths.ToString().ToLower()}");
+      sb.Append ($", restoreScroll: {configuration.EnableScrollToInitialPosition.ToString().ToLower()}");
+      sb.Append (", runOnly: {type:\"tag\",values:[");
+      sb.Append ($"\"{AccessibilityConformanceLevelConverter.ConvertToString (configuration.ConformanceLevel)}\"");
+
+      if (configuration.ConformanceLevel == AccessibilityConformanceLevel.Wcag20_ConformanceLevelDoubleA)
+        AppendWcag2A (sb);
+
+      //axe version 3.2.2 requires the experimental tag for WCAG 2.1 A and WCAG 2.1 AA
+      if (configuration.ConformanceLevel == AccessibilityConformanceLevel.Wcag21_ConformanceLevelA)
```
MichaelKetting: Test missing?

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AccessibilityConfigurationJsonSerializer.cs
```diff
@@ -0,0 +1,48 @@
...
+      {
+        AppendWcag2A (sb);
+        AppendExperimental (sb);
+      }
+
```
MichaelKetting: Case WCAG 2.0 A missing?

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AccessibilityResultMapper.cs
```diff
@@ -0,0 +1,135 @@
...
+    {
+      return ruleTags
+          .Where (t => !AccessibilityConformanceLevelConverter.IsValid (t))
+          .Select (ParseAccessibilityTestSuccessCriteria);
+    }
+
+    private AccessibilityTestSuccessCriteria ParseAccessibilityTestSuccessCriteria (string rawTag)
+    {
+      var tagWithReplacedWCAG = Regex.Replace (rawTag, @"^wcag(\d)(\d)(\d{1,2})$", "wcag_$1_$2_$3");
+      var tagWithReplacedWCAGAndSeparators = Regex.Replace (tagWithReplacedWCAG, @"(\.|-)", "_");
```
MichaelKetting: Testcase missing?
Also, I thought that this one is just with section508? Should the regex be more specific?

Remotion/ObjectBinding/Web.IntegrationTests/BocEnumValue/AccessibilityBocEnumValueTest.cs
Stefan-Ilic: The test named "Validate" was dropped because it was just a copy-pasta of another test without even calling `Validate`, it was an oversight.


I went over all the Boc tests again, the following controls are missing validator tests:

* AutocompleteReferenceValue
* BooleanValue
* Checkbox
	* Is it even possible to trigger a validation error with this control?
* DateTimeValue
* List
* ListAsGrid
* ReferenceValue
* TreeView
	* Is it even possible to trigger a validation error with this control?

Meaning the following controls do have a validator test:

* EnumValue
* TextValue
* MultiLineTextValue

I will add the missing validator tests now

Remotion/ObjectBinding/Web.IntegrationTests/BocList/AccessibilityBocListTest.cs
Stefan-Ilic: Nice catch ⚾ The `Empyt` BocList on our test site is the one with placeholder text. I've added another one without the placeholder and sure enough, it fails the "AriaRequiredChildren" rule

Remotion/ObjectBinding/Web.IntegrationTests/BocEnumValue/AccessibilityBocEnumValueTest.cs
MichaelKetting: TreeView is readOnly, so, no.
BocCheckbox: you could probably build something with a custom validator. Probably easier to show than tell.

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AccessibilityResultMapper.cs
Stefan-Ilic: I verified that the second regex is indeed also for aXe stuff like `cat.structure` and not only for section508. Also, I adapted a unit test to also map one of those tags

Remotion/Web/Development.WebTesting/Accessibility/Implementation/AccessibilityResultMapper.cs
```diff
@@ -0,0 +1,160 @@
...
+          .Where (t => !AccessibilityConformanceLevelConverter.IsValid (t))
+          .Where (t => !_wellKnownSuccessCriteriaToIgnore.Contains (t))
+          .Select (ParseAccessibilityTestSuccessCriteria);
+    }
+
+    private AccessibilityTestSuccessCriteria ParseAccessibilityTestSuccessCriteria (string rawTag)
+    {
+      var transformedTag = Regex.Replace (rawTag, @"^wcag(\d)(\d)(\d{1,2})$", "wcag_$1_$2_$3");
+
+      if (transformedTag == rawTag)
```
MichaelKetting: Den guard können wir uns sparen.

Remotion/ObjectBinding/Web.IntegrationTests/BocBooleanValue/AccessibilityBocBooleanValueTest.cs
```diff
@@ -0,0 +1,109 @@
...
+      var bocBooleanValue = home.BooleanValues().GetByLocalID ("DeceasedField_Normal");
+      var analyzer = Helper.CreateAccessibilityAnalyzer();
+
+      var result = bocBooleanValue.Analyze (analyzer);
+
+      Assert.That (result.Violations, Is.Empty);
+    }
+
+    [Test]
+    public void NormalAndUnitialized_Validate ()
```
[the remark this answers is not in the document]
Stefan-Ilic: Ah, of course it works if I manually click the validate button, I tried calling `GetValidationErrors` and checking if it's not empty in my first draft of the commit and then replaced it with clicking the validate button, without ever doing both 😅

Remotion/ObjectBinding/Web.Development.WebTesting.TestSite/Controls/BocListAsGridUserControl.ascx.cs
```diff
@@ -49,6 +50,9 @@ protected override void OnInit (EventArgs e)
       JobList_Normal.EditableRowChangesCanceled += EditableRowChangesCanceledHandler;
       JobList_Normal.ListItemCommandClick += ListItemCommandClickHandler;
       JobList_Normal.CustomCellClick += CustomCellClickHandler;
+
+      AlwaysInvalidValidator.ServerValidate += (source, args) => args.IsValid = false;
+      JobList_AlwaysInvalid.RegisterValidator (AlwaysInvalidValidator);
```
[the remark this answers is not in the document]
Stefan-Ilic: If I try to reference the control in the view via `ControlToValidate` I get a

```
HttpException (0x80004005): Control 'JobList_AlwaysInvalid' referenced by the ControlToValidate property of 'AlwaysInvalidValidator' cannot be validated.]
```

I just thought this is a BocList quirk and left it at that. I could try and investigate but it didn't seem necessary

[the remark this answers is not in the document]
Stefan-Ilic: I also get red squigglies at compile time so obviously the compiler knows something I don't

Remotion/ObjectBinding/Web.IntegrationTests/BocList/AccessibilityBocListTest.cs
```diff
@@ -0,0 +1,163 @@
...
+    {
+      var home = Start();
+      var bocList = home.Lists().GetByLocalID ("JobList_Empty_WithoutPlaceholder");
+      var analyzer = Helper.CreateAccessibilityAnalyzer();
+
+      var result = bocList.Analyze (analyzer);
+
+      var filteredViolations = result.Violations.IgnoreByRuleIDAndXPath (
+              AccessibilityRuleID.AriaRequiredChildren,
+              "/div[@id='body_DataEditControl_JobList_Empty_WithoutPlaceholder_TableScrollContainer']/table/tbody")
```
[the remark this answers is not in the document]
Stefan-Ilic: I have compiled a condensed list of all tests that need to ignore violations to be green. We should probably look them over together to decide what gets a helper filter method and what doesn't

PR #37, opened 2019-12-04
RM-7309: Add workaround for exceptions not showing up in VS

iam-mholle:


iam-mholle:
Added to master with 457e0781aed7196a3245754818dc0183a26519ac.

PR #38, opened 2019-12-05
RM-7199: Escape search term before using it for highlighting

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7199

### DIFF
--- a/Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js
+++ b/Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js
@@ -1156,8 +1156,11 @@
                 if (!data[i])
                     continue;
                 var item = options.formatItem(data[i].data, i + 1, max, data[i].value, term);
+                var termElement = document.createElement("div");
+                termElement.innerText = term;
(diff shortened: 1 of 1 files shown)


Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js
```diff
@@ -1157,7 +1157,12 @@
                     continue;
                 var item = options.formatItem(data[i].data, i + 1, max, data[i].value, term);
                 var li = $ ("<li role='option' aria-selected='false' />")
-                  .html (options.highlight (item.html, term))
+                  .html (options.highlight (item.html, term
+                      .replace (/&/g, "&amp;")
```
iam-mholle: Not sure if this is the right place or if this should be in the highlight function

```javascript
highlight: function(value, term) {
    if (term == '')
      return value;
    return value.replace(new RegExp("(?![^&;]+;)(?!<[^<>]*)(" + term.replace(/([\^\$\(\)\[\]\{\}\*\.\+\?\|\\])/gi, "\\$1") + ")(?![^<>]*>)(?![^&;]+;)", "gi"), "<strong>$1</strong>");
},
```

PR #45, opened 2020-01-07
RM-7316: The Firefox CommandTimeout is set to 10 seconds and not affected by the CommandTimeout given in the DriverConfiguration

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7316


iam-mholle:
Added to master with 04f034531e6a775742efe1b6a526614f9808fb9a.

PR #46, opened 2020-01-13
RM-7320: Flaky Test: ScreenshotTest.BadgeTest in Firefox

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7320

### DIFF
--- a/Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/FirefoxBrowserContentLocator.cs
+++ b/Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/FirefoxBrowserContentLocator.cs
@@ -1,4 +1,4 @@
-// This file is part of the re-motion Core Framework (www.re-motion.org)
+﻿// This file is part of the re-motion Core Framework (www.re-motion.org)
 // Copyright (c) rubicon IT GmbH, www.rubicon.eu
 // 
 // The re-motion Core Framework is free software; you can redistribute it 
@@ -65,7 +65,7 @@ namespace Remotion.Web.Development.WebTesting.ScreenshotCreation.BrowserContentL
     [NotNull]
     private AutomationElement GetFirefoxContentElement (AutomationElement firefoxWindow)
     {
-      for (var i = 0; i < 3; i++)
+      for (var i = 0; i < 11; i++)
       {
         try
         {
@@ -78,7 +78,7 @@ namespace Remotion.Web.Development.WebTesting.ScreenshotCreation.BrowserContentL
         {
         }
 
-        Thread.Sleep (TimeSpan.FromSeconds (1));
+        Thread.Sleep (TimeSpan.FromMilliseconds (200));
       }
 
       throw new InvalidOperationException ("The Firefox content window could not be found within a timeout of 2 seconds.");
--- a/Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/FirefoxBrowserContentLocator.cs
+++ b/Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/FirefoxBrowserContentLocator.cs
@@ -17,7 +17,9 @@
 using System;
 using System.Drawing;
 using System.Linq;
+using System.Threading;
 using System.Windows.Automation;
+using JetBrains.Annotations;
 using OpenQA.Selenium;
 using Remotion.Utilities;
 
@@ -60,13 +62,36 @@ namespace Remotion.Web.Development.WebTesting.ScreenshotCreation.BrowserContentL
       return result;
     }
 
+    [NotNull]
     private static AutomationElement GetFirefoxContentElement (AutomationElement firefoxWindow)
+    {
+      foreach (var _ in Enumerable.Range (0, 3))
+      {
+        try
+        {
+          var result = GetFirefoxDocumentControl (firefoxWindow);
+
+          if (result != null)
+            return result;
+        }
+        catch (ElementNotAvailableException)
+        {
+        }
+
+        Thread.Sleep (TimeSpan.FromSeconds (1));
+      }
+
+      throw new TimeoutException ("The Firefox content window could not be found within a timeout of 2 seconds.");
+    }
+
+    [CanBeNull]
+    private static AutomationElement GetFirefoxDocumentControl (AutomationElement firefoxWindow)
     {
       return firefoxWindow.FindFirst (
           TreeScope.Subtree,
           new PropertyCondition (
-              AutomationElement.LocalizedControlTypeProperty,
-              "document"));
+              AutomationElement.ControlTypeProperty,
+              ControlType.Document));
     }
 
     private Tuple<int, AutomationElement> RateAutomationElement (AutomationElement window, IWebDriver driver)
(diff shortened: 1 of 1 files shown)


Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/FirefoxBrowserContentLocator.cs
```diff
@@ -60,13 +62,36 @@ public Rectangle GetBrowserContentBounds (IWebDriver driver)
       return result;
     }
 
+    [NotNull]
     private static AutomationElement GetFirefoxContentElement (AutomationElement firefoxWindow)
+    {
+      foreach (var _ in Enumerable.Range (0, 3))
```
Stefan-Ilic: I started wondering again why we began using Enumerable.Range instead of a traditional for loop and came across [this](https://stackoverflow.com/questions/915745/thoughts-on-foreach-with-enumerable-range-vs-traditional-for-loop) discussion. I'm interested in @MichaelKetting 's opinion

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/FirefoxBrowserContentLocator.cs
```diff
@@ -60,13 +62,36 @@ public Rectangle GetBrowserContentBounds (IWebDriver driver)
...
+        }
+
+        Thread.Sleep (TimeSpan.FromSeconds (1));
+      }
+
+      throw new TimeoutException ("The Firefox content window could not be found within a timeout of 2 seconds.");
+    }
+
+    [CanBeNull]
+    private static AutomationElement GetFirefoxDocumentControl (AutomationElement firefoxWindow)
```
Stefan-Ilic: I believe we're supposed to avoid static

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/FirefoxBrowserContentLocator.cs
```diff
@@ -60,13 +62,36 @@ public Rectangle GetBrowserContentBounds (IWebDriver driver)
...
+            return result;
+        }
+        catch (ElementNotAvailableException)
+        {
+        }
+
+        Thread.Sleep (TimeSpan.FromSeconds (1));
+      }
+
+      throw new TimeoutException ("The Firefox content window could not be found within a timeout of 2 seconds.");
```
Stefan-Ilic: Actually, after looking at this again, I believe the main selling point is the retry logic and not the timeout logic. I'm actually more in favor of an `InvalidOperationException` that goes like "...could not be found after 3 retries." or something like that. Thoughts?

Stefan-Ilic: Just to make this clear: To me, it doesn't feel like a timeout, but like a retry that happens to have a timeout and was wondering what your thoughts are on that

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/FirefoxBrowserContentLocator.cs
iam-mholle: Imo it makes a loop where the index variable is unused easier to understand. A traditional for loop has much unneeded clutter around the range, which is the important part here - think of the range concept in other languages (pseudo: `for _ in 0..3`).

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/FirefoxBrowserContentLocator.cs
iam-mholle: Since the calling method in this class has no way to configure either the timeout or the retry count, the only measurable effect is the timeout, thus the `TimeoutException`. Considering the use in the web tests, the method is used to get coordinates for mostly screenshot operations - the tester using this functionality might be better off seeing the `InvalidOperationException`.

So you are right, this probably should be an `InvalidOperationException`.

Interested in @MichaelKetting 's opinion.

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/FirefoxBrowserContentLocator.cs
iam-mholle: Right, I'll change both methods.

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/FirefoxBrowserContentLocator.cs
MichaelKetting: TimeoutException.

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/FirefoxBrowserContentLocator.cs
MichaelKetting: hmm...since i grew up with C and Assembler, I'm used to seeing a for-loop in this case and wouldn't wonder about things when I see it. Enumerable.Range to me feels like extra overhead when reasoning on the code since it's not somehting that's in my normal vocabulary. Enumable.Range is just such a special case feature... Probably a lot to do with personal preferences.

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/FirefoxBrowserContentLocator.cs
MichaelKetting: Yep, don't use static unless you want to communicate something with it (or need it because it's called from another static member).

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/FirefoxBrowserContentLocator.cs
MichaelKetting: Erm, just read the first comment. Yes, we don't find the window within an acceptable timeframe (=Timeout) but instead, we don't find the window and just stop when it's appropriate (=invalidopex)

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/FirefoxBrowserContentLocator.cs
```diff
@@ -60,13 +62,36 @@ public Rectangle GetBrowserContentBounds (IWebDriver driver)
       return result;
     }
 
-    private static AutomationElement GetFirefoxContentElement (AutomationElement firefoxWindow)
+    [NotNull]
+    private AutomationElement GetFirefoxContentElement (AutomationElement firefoxWindow)
```
MichaelKetting: summary comment for the why/how-to

iam-mholle:
Added to master with 82bec97a4f88afa8275eb594921fe1edc261e97d.

PR #47, opened 2020-01-15
RM-7191: Force browser scaling in webtests

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7191


MichaelKetting:
> I tested it locally and it seems to work. When I tried to use my installed Edge I got a weird error, meaning a newer Edge version broke something, but that's a different task entirely. Good job with this one 👍
@Stefan-Ilic since it broke something for you, did you already create a ticket for this?

iam-mholle:
Added to master with 21969dbb49af67e92d4dcb388f0404058e17193c.

Stefan-Ilic:
> > I tested it locally and it seems to work. When I tried to use my installed Edge I got a weird error, meaning a newer Edge version broke something, but that's a different task entirely. Good job with this one 👍
> > @Stefan-Ilic since it broke something for you, did you already create a ticket for this?

It's one of my post-its, will do today

PR #48, opened 2020-01-20
RM-7324: Update DependDB packages

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7324


iam-mholle:
Added to master with 23fb6df35ce3e437d7fcb461c21681e3e9263926.

PR #49, opened 2020-01-20
RM-7325: WebTests fail to start with Edge v79 and matching msedgedriver

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7325


iam-mholle:
`WebTestConfigurationFactory.LatestTestedChromeVersion` should also be updated to 79 to keep the versions in sync.

iam-mholle:
Open questions:
- Could/should a caching decorator be added for `IBrowserContentLocator`'s? (separate issue)
- Whats the browser rating in `ChromeBrowserContentLocator` for?

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/ChromeBrowserContentLocator.cs
```diff
@@ -27,9 +27,9 @@
 namespace Remotion.Web.Development.WebTesting.ScreenshotCreation.BrowserContentLocators
```
Stefan-Ilic: I'm commenting this here because I can't comment on the changed reference screenshot directly. After changing ScreenshotTest_TableContainer.png the screenshot test now fails in Firefox (at least locally on my machine)

Chrome, Edge and even IE seemed fine

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/EdgeBrowserContentLocator.cs
```diff
@@ -0,0 +1,112 @@
+﻿using System;
```
Stefan-Ilic: Missing license header

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/EdgeBrowserContentLocator.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Threading;
+using System.Windows;
+using System.Windows.Automation;
+using JetBrains.Annotations;
+using OpenQA.Selenium;
+using Remotion.Web.Development.WebTesting.Utilities;
+
+namespace Remotion.Web.Development.WebTesting.ScreenshotCreation.BrowserContentLocators
+{
+  public class EdgeBrowserContentLocator : IBrowserContentLocator
```
Stefan-Ilic: Missing class docs

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/EdgeBrowserContentLocator.cs
```diff
@@ -0,0 +1,112 @@
...
+        throw new InvalidOperationException ("Could not resolve the bounds of the Edge browser window.");
+
+      return new Rectangle (
+          (int) Math.Round (rawBounds.X),
+          (int) Math.Round (rawBounds.Y),
+          (int) Math.Round (rawBounds.Width),
+          (int) Math.Round (rawBounds.Height));
+    }
+
+    private TResult RetryUntil<TResult> (Func<TResult> action, TResult defaultValue, int retries, TimeSpan interval)
```
Stefan-Ilic: Rename parameter "action" to "function" etc. ?

Stefan-Ilic: Also, what do you think about changing the name "defaultValue" to something like "valueToCompareTo" etc (to express that it's the value that actually triggers the retry if the result is that value)

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/EdgeBrowserContentLocator.cs
```diff
@@ -0,0 +1,112 @@
...
+      var executor = (IJavaScriptExecutor) driver;
+      var previousTitle = JavaScriptExecutor.ExecuteStatement<string> (executor, c_setWindowTitle, id);
+
+      var result = RetryUntil (
+          () => windows.SingleOrDefault (w => w.Current.Name.StartsWith (id)),
+          null,
+          3,
+          TimeSpan.FromMilliseconds (100));
+
+      JavaScriptExecutor.ExecuteStatement<string> (executor, c_setWindowTitle, previousTitle);
```
Stefan-Ilic: Would there be a benefit in putting the "change windowtitle back logic" in a finally block?
I personally can't think of one since it's just for the duration of the test but who knows

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/EdgeBrowserContentLocator.cs
```diff
@@ -0,0 +1,112 @@
...
+                  new PropertyCondition (AutomationElement.ControlTypeProperty, ControlType.Document),
+                  new PropertyCondition (AutomationElement.FrameworkIdProperty, "Chrome"))),
+          null,
+          5,
+          TimeSpan.Zero);
+
+      if (element == null)
+        throw new InvalidOperationException ("Could not find the content window of the found Edge browser window.");
+
+      var rawBounds = RetryUntil (() => element.Current.BoundingRectangle, Rect.Empty, 3, TimeSpan.FromMilliseconds (100));
```
Stefan-Ilic: Is the retry really necessary at this point? Since the object is already there and its `BoundingRectangle` property shouldn't change anymore, unless it is evaluated lazily?

iam-mholle: It is necessary as the value is evaluated upon accessing it.

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/ChromeBrowserContentLocator.cs
iam-mholle: The failing test must be caused by your setup, as Firefox uses `Firefox.any.ScreenshotTest_TableContainer.png` as reference image.

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/EdgeBrowserContentLocator.cs
iam-mholle: The `defaultValue` is the value that the `Func` result is compared with, and the value that is returned if the timeout is exceeded. Imo `defaultValue` expresses this good enough, I'll leave as-is until the lead review.

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/EdgeBrowserContentLocator.cs
iam-mholle: No exceptions are to be expected after calling `JavaScriptExecutor.ExecuteStatement`, I see no benefit. Will leave open.

iam-mholle:
There seems to be a cleanup issue with the new Chrome version.

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/EdgeBrowserContentLocator.cs
Stefan-Ilic: Actually, the other content locators don't use inheritdoc and specify a new doc here.

iam-mholle: Thanks, fixed.

Remotion/Web/Development.WebTesting/BrowserSession/BrowserSessionBase.cs
```diff
@@ -35,7 +35,7 @@ public abstract class BrowserSessionBase<T> : IBrowserSession
   {
     private const int c_driverShutdownWaitTime = 250;
     private const int c_browserShutdownWaitTime = 500;
-    private const int c_browserSubProcessShutdownWaitTime = 100;
+    private const int c_browserSubProcessShutdownWaitTime = 250;
```
MichaelKetting: Should consider changing the times to TimeSpan, otherwise we don't know what the value means. Appending the unit is also possible, but not preferred. -> separate task

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/EdgeBrowserContentLocator.cs
MichaelKetting: > No exceptions are to be expected after calling

Is this something that's inherrently flawed/impossible or just a best practice we expect people to follow? Typically, a reliable cleaning up is always the first choice unless it's a) expensive (code or runtime wise) and b) reasonably likely to be irrelevant.

Just regarding the finally-block itself: if we can expect there to be no excepton, yes, we could probably skip it, again, most of the time, it's not worth it to consider the viability of not using a finally-block conpared to just doing it. Please also consider future maintainer of the code: without the finally block they are requried to a) notice there is a cleanup semantic and b) not introduce code that might throw while there is no finally block.

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/EdgeBrowserContentLocator.cs
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    private Rectangle ResolveBoundsFromWindow (AutomationElement window)
+    {
+      var element = RetryUntil (
+          () => window.FindFirst (
+              TreeScope.Subtree,
+              new AndCondition (
+                  new PropertyCondition (AutomationElement.ControlTypeProperty, ControlType.Document),
+                  new PropertyCondition (AutomationElement.FrameworkIdProperty, "Chrome"))),
```
MichaelKetting: So, Edge uses "Chrome" here? Should make this more explicit to show that this isn't a mistake.

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/EdgeBrowserContentLocator.cs
```diff
@@ -0,0 +1,131 @@
...
+using JetBrains.Annotations;
+using OpenQA.Selenium;
+using Remotion.Web.Development.WebTesting.Utilities;
+
+namespace Remotion.Web.Development.WebTesting.ScreenshotCreation.BrowserContentLocators
+{
+  /// <summary>
+  /// Locates the browser content area for Edge.
+  /// </summary>
+  public class EdgeBrowserContentLocator : IBrowserContentLocator
```
MichaelKetting: How much of this logic is identical to the ChromeBrowserContentLocator.

Disclaimer: did not check if the window finding a) works and b) could be simplified. Am assuming this was already reviewed / forked from exisitng content locators :)

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/EdgeBrowserContentLocator.cs
iam-mholle: `var previousTitle = JavaScriptExecutor.ExecuteStatement<string> (executor, c_setWindowTitle, id);` (l.73) is the only place where an exception could happen, we're not expecting it tho. Adding a `try..finally` block calling the same method again does not seem ideal. But again, we don't expect it not to work.

MichaelKetting: No, the idea would be to put the 2nd executeStatement into the finally block and the rest of the method after the 1st executeStatement into the try-block

iam-mholle: Then imo `b) reasonably likely to be irrelevant` is the case.

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/EdgeBrowserContentLocator.cs
iam-mholle: Not a significant amount. I intentionally dropped the browser window rating, as I could not see benefits. The automation tree structure also looks different now, therefore the logic had to be adapted.

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/EdgeBrowserContentLocator.cs
iam-mholle: Yes, I'll make it more explicit.
Fun fact: Edge now looks like an electron window through the automation api - Chrome doesn't.

Remotion/Web/Development.WebTesting/BrowserSession/BrowserSessionBase.cs
iam-mholle: Added https://github.com/re-motion/Trunk-SVN2Git/pull/49/commits/26b847a98c6f4b15888032133938af87427fe440.

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/EdgeBrowserContentLocator.cs
iam-mholle: Adding `try..finally` to express intent.

Remotion/Web/Development.WebTesting/BrowserSession/BrowserSessionBase.cs
```diff
@@ -121,13 +121,13 @@ public virtual void Dispose ()
       // Check driver and main browser for null
       if (driverProcess != null)
       {
-        ProcessUtils.GracefulProcessShutdown (driverProcess, c_driverShutdownWaitTime);
+        ProcessUtils.GracefulProcessShutdown (driverProcess, (int) _driverShutdownWaitTime.TotalMilliseconds);
```
MichaelKetting: Change signature of GracefulProcessShutdown to use TImespan.

Remotion/Web/Development.WebTesting/BrowserSession/BrowserSessionBase.cs
```diff
@@ -121,13 +121,13 @@ public virtual void Dispose ()
       // Check driver and main browser for null
       if (driverProcess != null)
       {
-        ProcessUtils.GracefulProcessShutdown (driverProcess, c_driverShutdownWaitTime);
+        ProcessUtils.GracefulProcessShutdown (driverProcess, (int) _driverShutdownWaitTime.TotalMilliseconds);
         driverProcess.WaitForExit (60000);
```
MichaelKetting: Why do we have that if we use GracefulProcessShutdown?
Yes, this might be a separate issue.

iam-mholle: Added [RM-7348](https://re-motion.atlassian.net/browse/RM-7348).

Remotion/Web/Development.WebTesting/BrowserSession/BrowserSessionBase.cs
```diff
@@ -137,7 +137,7 @@ private void CheckForSubProcessesExited (List<Process> browserSubProcesses, Proc
       // Wait for every sub process of the browser process to be closed
       for (int i = 0; i < 5; i++)
       {
-        if (WaitForSubProcessExit (browserSubProcesses, c_browserSubProcessShutdownWaitTime))
+        if (WaitForSubProcessExit (browserSubProcesses, (int) _browserSubProcessShutdownWaitTime.TotalMilliseconds))
```
MichaelKetting: Sorry, found one more occurance. this should be it, now.

iam-mholle:
Added to master with 1641ceb84c8256425608cc9635f9b73b7368ed3d.

PR #55, opened 2020-02-27
RM-7319: ResolveBorderElementA retry loop

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7319

### DIFF
--- a/Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.cs
+++ b/Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.cs
@@ -675,11 +675,7 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests
         {
           action();
         }
-        catch (AssertionException) when (i + 1 == retries)
-        {
-          throw;
-        }
-        catch (AssertionException)
+        catch (AssertionException) when (i + 1 < retries)
         {
           continue;
         }
--- a/Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.cs
+++ b/Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.cs
@@ -371,7 +371,7 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests
 
             Helper.RunScreenshotTestExact<IFluentScreenshotElement<ElementScope>, ScreenshotTest> (element, ScreenshotTestingType.Both, test);
           },
-          3);
+          2);
     }
 
     [Category ("Screenshot")]
@@ -669,13 +669,16 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests
 
     private void RetryTest (Action action, int retries)
     {
-      for (int i = 0; i < retries; i++)
+      if (retries < 0)
+        throw new ArgumentOutOfRangeException ("retries", "Retries must be greater than or equal to zero.");
+
+      for (int i = 0; i <= retries; i++)
       {
         try
         {
           action();
         }
-        catch (AssertionException) when (i + 1 < retries)
+        catch (AssertionException) when (i < retries)
         {
           continue;
         }
--- a/Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.cs
+++ b/Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.cs
@@ -681,7 +681,10 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests
         }
         catch (AssertionException)
         {
+          continue;
         }
+
+        return;
       }
     }
 
--- a/Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.cs
+++ b/Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.cs
@@ -358,14 +358,19 @@ namespace Remotion.Web.Development.WebTesting.IntegrationTests
     [Test]
     public void ResolveBorderElementA ()
     {
-      var home = Start();
+      RetryTest (
+          () =>
+          {
+            var home = Start();
 
-      ScreenshotTestingDelegate<IFluentScreenshotElement<ElementScope>> test =
-          (builder, target) => { builder.Crop (target, new WebPadding (1)); };
+            ScreenshotTestingDelegate<IFluentScreenshotElement<ElementScope>> test =
+                (builder, target) => { builder.Crop (target, new WebPadding (1)); };
 
-      var element = home.Scope.FindId ("borderElementA").ForElementScopeScreenshot();
+            var element = home.Scope.FindId ("borderElementA").ForElementScopeScreenshot();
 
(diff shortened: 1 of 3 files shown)


Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.cs
```diff
@@ -358,14 +358,19 @@ public void ResolveAbsolutePositionedElement ()
...
+            ScreenshotTestingDelegate<IFluentScreenshotElement<ElementScope>> test =
+                (builder, target) => { builder.Crop (target, new WebPadding (1)); };
 
-      var element = home.Scope.FindId ("borderElementA").ForElementScopeScreenshot();
+            var element = home.Scope.FindId ("borderElementA").ForElementScopeScreenshot();
 
-      Helper.RunScreenshotTestExact<IFluentScreenshotElement<ElementScope>, ScreenshotTest> (element, ScreenshotTestingType.Both, test);
+            Helper.RunScreenshotTestExact<IFluentScreenshotElement<ElementScope>, ScreenshotTest> (element, ScreenshotTestingType.Both, test);
+          },
+          3);
```
Stefan-Ilic: I'm guessing that 3 was the agreed upon amount of retries?

Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.cs
```diff
@@ -661,6 +666,32 @@ public void ScreenshotTransformationTest ()
       Helper.RunScreenshotTestExact<ScreenshotTest> (PrepareTest(), ScreenshotTestingType.Browser, test);
     }
 
+    private void RetryTest (Action action, int times)
```
Stefan-Ilic: Do we want to make a //todo for removing this helper method once NUnit 3 gives us the option to retry test runs?

Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.cs
iam-mholle: Not yet. I do think that 3 would be sufficient.

Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.cs
iam-mholle: Added todo comment

Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.cs
```diff
@@ -661,6 +667,32 @@ public void ScreenshotTransformationTest ()
...
```
MichaelKetting: "times" => "londonTimes" or "newYorkTimes"?

Yes, we need a better name.

Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.cs
```diff
@@ -661,6 +667,32 @@ public void ScreenshotTransformationTest ()
...
+          continue;
+        }
+
+        lastThrownException = null;
+        break;
+      }
+
+      if (lastThrownException != null)
+      {
+        throw lastThrownException;
```
MichaelKetting: doing it this way kills our stacktrace. code needs to be rewritten to rethrow immediately if the max number of tries is exceeded. in theory, we could also do a when-clause on the catch block. Otherwise, it's just an IF and a THROW w/o the exception variable.

Remotion/Web/Development.WebTesting/ScreenshotCreation/ScreenshotBuilder.cs
```diff
@@ -130,9 +133,14 @@ public void Save ([NotNull] string path, bool @override = false)
     {
       ArgumentUtility.CheckNotNullOrEmpty ("path", path);
 
-      if (!@override && File.Exists (path))
+      var fileExists = File.Exists (path);
```
MichaelKetting: isFileExisting

Remotion/Web/Development.WebTesting/ScreenshotCreation/ScreenshotBuilder.cs
```diff
@@ -130,9 +133,14 @@ public void Save ([NotNull] string path, bool @override = false)
     {
       ArgumentUtility.CheckNotNullOrEmpty ("path", path);
 
-      if (!@override && File.Exists (path))
+      var fileExists = File.Exists (path);
+
+      if (!@override && fileExists)
         throw new InvalidOperationException (string.Format ("A screenshot with the file name '{0}' does already exist.", path));
 
+      if (@override && fileExists)
```
MichaelKetting: ReSharper probably already mentioned that the override flag is redundant.

Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.cs
```diff
@@ -661,6 +667,27 @@ public void ScreenshotTransformationTest ()
...
 
+    private void RetryTest (Action action, int retries)
+    {
+      for (int i = 0; i < retries; i++)
+      {
+        try
+        {
+          action();
+        }
+        catch (AssertionException) when (i + 1 == retries)
```
MichaelKetting: can we change this to use the when only to trigger the continue and not include the catch/throw-block at all? The way it's now does work but it seems overly complex.

Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.cs
```diff
@@ -661,6 +667,23 @@ public void ScreenshotTransformationTest ()
...
 
+    private void RetryTest (Action action, int retries)
+    {
+      for (int i = 0; i < retries; i++)
+      {
+        try
+        {
+          action();
+        }
+        catch (AssertionException) when (i + 1 < retries)
```
MichaelKetting: Hmm...is the condition wrong:
retries = 0: never continues because i=0, 0+1 < 0 == false
retries = 1: never continues because i=0, 0+1 < 1 == false
retries = 2: continues 1x because i=1, 1+1 < 2 == false

iam-mholle: Sorry the local

iam-mholle: Sorry I thought of the `retries` parameter as **tries** (total = `retries`) rather than **retries** (total = `1 + retries`) - this way, the condition makes sense to me, the last occurring `AssertionException` is not caught.

Which concept should I choose?
1. parameter indicates total tries
2. parameter indicates retries after an initial try

MichaelKetting: "retries". otherwise, we could pass tries=0 and never do anything :p

iam-mholle:
Added to master with 15b7d777ec3dc79db0644d9ea709585e51dfb325.

PR #62, opened 2020-03-23
RM-7387: Edge cannot find alert dialogs - workaround

Stefan-Ilic:
https://re-motion.atlassian.net/browse/RM-7387


Stefan-Ilic:
Added to `master` with e3ff1335132f2dccc1ec605931498ababa7372f9

PR #66, opened 2020-03-27
RM-7393: Add ability to configure AsyncJavascriptTimeout in DriverConfiguration

Stefan-Ilic:
### Jira Issue
https://re-motion.atlassian.net/browse/RM-7393

### Breaking Changes
DriverConfiguration now requires an additional `TimeSpan` parameter in its constructor for the `asyncJavaScriptTimeout`.

### How to test
Automated tests have been provided. In addition, to test if this actually does something in real scenario you can run `AccessibilityIFrameSupportTest.TestInjectSourceToIFrames` inside Internet Explorer (Yes, IE, sorry). It should fail before this PR and succeed with these changes.

### Checklist

- [x] I have adapted schema files if I made changes to XML Config Files.
- [x] All new files have license headers.
- [x] All new public classes, methods and properties have XML documentation and JetBrains annotations.
- [x] I have identified all breaking changes and marked them appropriately.
- [x] I have followed the style guide to the best of my ability.
- [x] If I added new projects to the solution, I also added them to the corresponding `projects.props` file.

### DIFF
--- a/Remotion/Web/Development.WebTesting/Configuration/WebTestConfigurationSection.cs
+++ b/Remotion/Web/Development.WebTesting/Configuration/WebTestConfigurationSection.cs
@@ -17,6 +17,7 @@
 using System;
 using System.Configuration;
 using System.IO;
+using OpenQA.Selenium;
 using Remotion.Utilities;
 using Remotion.Web.Development.WebTesting.DownloadInfrastructure;
 
@@ -33,6 +34,7 @@ namespace Remotion.Web.Development.WebTesting.Configuration
     private readonly ConfigurationProperty _browserProperty;
     private readonly ConfigurationProperty _searchTimeoutProperty;
     private readonly ConfigurationProperty _commandTimeoutProperty;
+    private readonly ConfigurationProperty _asyncJavaScriptTimeoutProperty;
     private readonly ConfigurationProperty _downloadStartedTimeoutProperty;
     private readonly ConfigurationProperty _downloadUpdatedTimeoutProperty;
     private readonly ConfigurationProperty _verifyWebApplicationStartedTimeoutProperty;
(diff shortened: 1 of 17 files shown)


Remotion/Web/Development.WebTesting/Configuration/WebTestConfigurationSection.cs
```diff
@@ -152,6 +156,14 @@ public TimeSpan CommandTimeout
       get { return (TimeSpan) this [_commandTimeoutProperty]; }
     }
 
+    /// <summary>
+    /// Specifies how long Selenium should maximally wait for asynchronous callbacks after calling <see cref="IJavaScriptExecutor.ExecuteAsyncScript" />  before failing.
```
iam-mholle:
```suggestion
    /// Specifies how long Selenium should maximally wait for asynchronous callbacks after calling <see cref="IJavaScriptExecutor.ExecuteAsyncScript" /> before failing.
```

Remotion/Web/Development.WebTesting/WebDriver/Factories/Chrome/ChromeBrowserFactory.cs
```diff
@@ -56,6 +56,7 @@ public IBrowserSession CreateBrowser (DriverConfiguration configuration)
       var commandTimeout = configuration.CommandTimeout;
 
       var driver = CreateChromeDriver (out var driverProcessID, commandTimeout);
+      driver.Manage().Timeouts().AsynchronousJavaScript = configuration.AsyncJavaScriptTimeout;
```
iam-mholle: IMO it's fine to set it here, but this could also be changed in the tests via some extension method for example 😀.

Stefan-Ilic: You're probably right but I do think doing it like this is cleaner

Stefan-Ilic:
Added to `master` with 03c9f59e246901e6cf3cd6e03392dbd0eef9ce2d

PR #70, opened 2020-04-15
RM-7405: Update Build Script to v2.0.0-alpha022

Stefan-Ilic:
### Jira Issue
https://re-motion.atlassian.net/browse/RM-7405

### Breaking Changes
None

### How to test
If all configurations were adapted correctly, there should be no difference between running builds and this branch and on develelop.

Due to a bug in NUnit 2.7.1, however, all tests marked as `[Explicit]` are not skipped and are being run instead. This bug is gone in NUnit 3.12.0.

### Checklist

- [x] I have adapted schema files if I made changes to XML Config Files.
- [x] All new files have license headers.
- [x] All new public classes, methods and properties have XML documentation and JetBrains annotations.
- [x] I have identified all breaking changes and marked them appropriately.
- [x] I have followed the style guide to the best of my ability.
- [x] If I added new projects to the solution, I also added them to the corresponding `projects.props` file.

### DIFF
--- a/Build-Remotion.cmd
+++ b/Build-Remotion.cmd
@@ -11,7 +11,7 @@ if not exist %msbuild% set msbuild="%program-path%\Microsoft Visual Studio\2017\
 set log-dir=build\BuildOutput\log
 set nuget-bin=build\BuildOutput\temp\nuget-bin
 set nuget=%nuget-bin%\nuget.exe
-set nuget-download=powershell.exe -NoProfile -Command "& {(New-Object System.Net.WebClient).DownloadFile('https://www.nuget.org/nuget.exe','%nuget%')}"
+set nuget-download=powershell.exe -NoProfile -Command "& {(New-Object System.Net.WebClient).DownloadFile('https://dist.nuget.org/win-x86-commandline/latest/nuget.exe','%nuget%')}"
 
 if not exist remotion.snk goto nosnk
 
--- a/Build/Shared.Server.Extensions.targets
+++ b/Build/Shared.Server.Extensions.targets
@@ -22,8 +22,10 @@
   <Target Name="ArtifactBuild_Configuration">
     <PropertyGroup>
       <Platforms>x64</Platforms>
-      <DatabaseSystems>SqlServer2012</DatabaseSystems>
-      <Browsers></Browsers>
+      <DatabaseSystems>SqlServer2012+NoDB</DatabaseSystems>
+      <Browsers>NoBrowser</Browsers>
+      <ExecutionRuntimes>Win_NET48</ExecutionRuntimes>
+      <TargetRuntimes>NET48</TargetRuntimes>
       <SkipDocumentation>True</SkipDocumentation>
     </PropertyGroup>
   </Target>
@@ -31,8 +33,10 @@
   <Target Name="CIProductionBuild_Configuration">
     <PropertyGroup>
       <Platforms>x64</Platforms>
-      <DatabaseSystems>SqlServer2012</DatabaseSystems>
-      <Browsers>Chrome</Browsers>
+      <DatabaseSystems>SqlServer2012+NoDB</DatabaseSystems>
+      <Browsers>Chrome+NoBrowser</Browsers>
+      <ExecutionRuntimes>Win_NET48+EnforcedLocalMachine</ExecutionRuntimes>
+      <TargetRuntimes>NET48</TargetRuntimes>
       <SkipDocumentation>True</SkipDocumentation>
     </PropertyGroup>
   </Target>
@@ -40,8 +44,10 @@
   <Target Name="FullBuildReduced_Configuration">
     <PropertyGroup>
       <Platforms>x64</Platforms>
-      <DatabaseSystems>SqlServer2012</DatabaseSystems>
-      <Browsers>Chrome+InternetExplorer</Browsers>
+      <DatabaseSystems>SqlServer2012+NoDB</DatabaseSystems>
+      <Browsers>Chrome+InternetExplorer+Firefox+Edge+NoBrowser</Browsers>
+      <ExecutionRuntimes>Win_NET48+EnforcedLocalMachine</ExecutionRuntimes>
+      <TargetRuntimes>NET48</TargetRuntimes>
       <TestCategoriesToExclude></TestCategoriesToExclude>
     </PropertyGroup>
   </Target>
--- a/Build/TestingSetupForWebTest.targets
+++ b/Build/TestingSetupForWebTest.targets
@@ -29,11 +29,14 @@
     <Error Text="The property 'Platform' is not set." Condition="'$(Platform)' == ''" />
     <Error Text="The property 'AppConfigFile' is not set." Condition="'$(AppConfigFile)' == ''" />
     <Error Text="The property 'Browser' is not set." Condition="'$(Browser)' == ''" />
+    <Error Text="The property 'HostTestSitesInDocker' is not set." Condition="'$(HostTestSitesInDocker)' == ''" />
 
     <PropertyGroup>
-      <_dockerEnabled>False</_dockerEnabled>
-      <_dockerEnabled Condition="'$(DisableDocker)' == 'False'">True</_dockerEnabled>
-      <_dockerImageName>mcr.microsoft.com/dotnet/framework/aspnet:4.6.2</_dockerImageName>
+      <_dockerImageTag Condition="'$(Browser)' == 'InternetExplorer'">4.8</_dockerImageTag>
+      <_dockerImageTag Condition="'$(Browser)' == 'Chrome'">4.8</_dockerImageTag>
+      <_dockerImageTag Condition="'$(Browser)' == 'Edge'">4.8</_dockerImageTag>
+      <_dockerImageTag Condition="'$(Browser)' == 'Firefox'">4.7.2</_dockerImageTag>
+      <_dockerImageName>mcr.microsoft.com/dotnet/framework/aspnet:$(_dockerImageTag)</_dockerImageName>
       <!-- Hostname cannot be longer than 63 characters-->
       <_dockerHostName>RemotionWebTestContainer</_dockerHostName>
       <_dockerPortNumber>60402</_dockerPortNumber>
@@ -42,12 +45,13 @@
       <_dockerVerifyWebApplicationStartedTimeout>00:01:00</_dockerVerifyWebApplicationStartedTimeout>
     </PropertyGroup>
 
+    <Error Text="No known browser could be identified." Condition="'$(_dockerImageTag)' == ''" />
     <Message Text="Selected Browser: '$(Browser)'" />
     <Message Text="The property 'ChromeVersionArchive' is set to '$(ChromeVersionArchive)'." />
     <Message Text="The property 'EdgeVersionArchive' is set to '$(EdgeVersionArchive)'." />
     <Message Text="The property 'FirefoxVersionArchive' is set to '$(FirefoxVersionArchive)'." />
-    <Message Text="Docker enabled: '$(_dockerEnabled)'" />
-    <Message Text="Using Docker image '$(_dockerImageName)' to host the test site." Condition="'$(_dockerEnabled)' == 'True'" />
+    <Message Text="Docker enabled: '$(HostTestSitesInDocker)'" />
+    <Message Text="Using Docker image '$(_dockerImageName)' to host the test site." Condition="'$(HostTestSitesInDocker)' == 'True'" />
 
     <MSBuild.ExtensionPack.Xml.XmlFile
         TaskAction="ReadAttribute"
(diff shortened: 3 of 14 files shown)


Build-Remotion.cmd
```diff
@@ -11,7 +11,7 @@ if not exist %msbuild% set msbuild="%program-path%\Microsoft Visual Studio\2017\
 set log-dir=build\BuildOutput\log
 set nuget-bin=build\BuildOutput\temp\nuget-bin
 set nuget=%nuget-bin%\nuget.exe
-set nuget-download=powershell.exe -NoProfile -Command "& {(New-Object System.Net.WebClient).DownloadFile('https://www.nuget.org/nuget.exe','%nuget%')}"
+set nuget-download=powershell.exe -NoProfile -Command "& {(New-Object System.Net.WebClient).DownloadFile('https://dist.nuget.org/win-x86-commandline/latest/nuget.exe','%nuget%')}
```
Stefan-Ilic, line 14: The new NUnit runner requires a newer version of nuget.exe, hence the new link

SecurityManager/Build/Remotion.Local.build
```diff
@@ -35,26 +35,37 @@
     <SkipSourceLinks>True</SkipSourceLinks>
     <SkipNuGet>True</SkipNuGet>
     <SkipDependDB>True</SkipDependDB>
-    <DisableDocker>True</DisableDocker>
+    <HostTestSitesInDocker>False</HostTestSitesInDocker>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <DockerImageDotNetFrameworkRuntime4_7 Condition="'$(DockerImageDotNetFrameworkRuntime4_7)' == ''">
+      Docker images are per default unavailable in local builds
```
Stefan-Ilic, line 43: Since `DockerImageDotNetFrameworkRuntimeXX` cannot be empty (by design) since TestConfigurations are validated against supported values even if they are not run, we need to set them to a non-empty value in local builds as well

Remotion/Build/Customizations/Projects.props
```diff
@@ -2,16 +2,51 @@
...
   </PropertyGroup>
 
   <PropertyGroup>
-    <TestingSetupBuildFileForDatabase>$(SolutionDirectory)Build\TestingSetupForDatabase.targets</TestingSetupBuildFileForDatabase>
-    <TestingSetupBuildFileForWebTest>$(SolutionDirectory)Build\TestingSetupForWebTest.targets</TestingSetupBuildFileForWebTest>
+    <WebtestingTestConfiguration Condition="'$(WebtestingTestConfiguration)' == ''">
+      Chrome           + NET472 + Debug   + x86 + NoDB + EnforcedLocalMachine;<!-- Upgrade to NET48 once it is supported -->
+      Firefox          + NET472 + Release + x64 + NoDB + EnforcedLocalMachine;
+      InternetExplorer + NET472 + Release + x64 + NoDB + EnforcedLocalMachine;<!-- Upgrade to NET48 once it is supported -->
+      Edge             + NET472 + Release + x64 + NoDB + EnforcedLocalMachine;<!-- Upgrade to NET48 once it is supported -->
```
Stefan-Ilic, line 14: Since these webtests are run in NET472 we currently cannot run them on agent 2 as long as it doesn't have .NET Framework 4.7.2 like the other agents

Build-Remotion.cmd
iam-mholle, line 14: Missing quote
```suggestion
set nuget-download=powershell.exe -NoProfile -Command "& {(New-Object System.Net.WebClient).DownloadFile('https://dist.nuget.org/win-x86-commandline/latest/nuget.exe','%nuget%')}"
```

SecurityManager/Build/Remotion.Server.build
```diff
@@ -50,17 +50,21 @@
...
   </Target>
   
   <Target Name="FullBuild_Configuration">
     <PropertyGroup>
       <Platforms>x86+x64</Platforms>
-      <DatabaseSystems>SqlServer2017+SqlServer2012+SqlServer2014+SqlServer2016</DatabaseSystems>
-      <Browsers>Chrome+InternetExplorer+Firefox+Edge</Browsers>
+      <ExecutionRuntimes>Win_NET472+Win_NET48+EnforcedLocalMachine</ExecutionRuntimes>
+      <DatabaseSystems>SqlServer2012+SqlServer2014+SqlServer2016+SqlServer2017</DatabaseSystems>
+      <Browsers>Chrome+InternetExplorer+Firefox+NoBrowser</Browsers>
```
MichaelKetting, line 66: What about Edge?

MichaelKetting: Or drop all the browsers?

Remotion/Build/Customizations/Projects.props
```diff
@@ -2,16 +2,51 @@
...
+    <NormalTestConfiguration Condition="'$(NormalTestConfiguration)' == ''">
+      Win_NET48  + NET48  + NoBrowser + NoDB + Debug   + x86;
+      Win_NET48  + NET48  + NoBrowser + NoDB + Release + x86;
+      Win_NET48  + NET48  + NoBrowser + NoDB + Debug   + x64;
+      Win_NET48  + NET48  + NoBrowser + NoDB + Release + x64;
+
+      <!-- Local-->
+      LocalMachine  + NET48  + NoBrowser + SqlServerDefault + Debug + x86;
+
+      <!-- Compatibility Mode -->
```
MichaelKetting, line 40: "compatibility mode"?

Remotion/Build/Customizations/Projects.props
```diff
@@ -2,16 +2,51 @@
...
-    <TestExecutionRuntime>net-4.6.1</TestExecutionRuntime>
+    <TestSetupBuildFileForDatabase>$(SolutionDirectory)Build\TestingSetupForDatabase.targets</TestSetupBuildFileForDatabase>
+    <TestSetupBuildFileForWebTest>$(SolutionDirectory)Build\TestingSetupForWebTest.targets</TestSetupBuildFileForWebTest>
   </PropertyGroup>
 
   <PropertyGroup>
-    <TestingSetupBuildFileForDatabase>$(SolutionDirectory)Build\TestingSetupForDatabase.targets</TestingSetupBuildFileForDatabase>
-    <TestingSetupBuildFileForWebTest>$(SolutionDirectory)Build\TestingSetupForWebTest.targets</TestingSetupBuildFileForWebTest>
+    <WebtestingTestConfiguration Condition="'$(WebtestingTestConfiguration)' == ''">
+      Chrome           + NET462 + Debug   + x86 + NoDB + EnforcedLocalMachine;<!-- Upgrade to NET48 once it is supported -->
```
MichaelKetting, line 11: remind me, how do we choose the docker image (.net 4.7.2 or .net 4.8) that is used to host IIS?

Build/TestingSetupForWebTest.targets
```diff
@@ -29,10 +29,9 @@
     <Error Text="The property 'Platform' is not set." Condition="'$(Platform)' == ''" />
     <Error Text="The property 'AppConfigFile' is not set." Condition="'$(AppConfigFile)' == ''" />
     <Error Text="The property 'Browser' is not set." Condition="'$(Browser)' == ''" />
+    <Error Text="The property 'HostTestSitesInDocker' is not set." Condition="'$(HostTestSitesInDocker)' == ''" />
 
     <PropertyGroup>
-      <_dockerEnabled>False</_dockerEnabled>
-      <_dockerEnabled Condition="'$(DisableDocker)' == 'False'">True</_dockerEnabled>
       <_dockerImageName>mcr.microsoft.com/dotnet/framework/aspnet:4.6.2</_dockerImageName>
```
MichaelKetting, line 35: okay, why do we use the 4.6.2 docker image for the docker container that hosts the iis?

Build/Shared.Server.Extensions.targets
```diff
@@ -22,26 +22,32 @@
...
     </PropertyGroup>
   </Target>
 
   <Target Name="FullBuildReduced_Configuration">
     <PropertyGroup>
       <Platforms>x64</Platforms>
-      <DatabaseSystems>SqlServer2012</DatabaseSystems>
-      <Browsers>Chrome+InternetExplorer</Browsers>
+      <DatabaseSystems>SqlServer2012+NoDB</DatabaseSystems>
+      <Browsers>Chrome+InternetExplorer+Firefox+Edge+NoBrowser</Browsers>
```
MichaelKetting, line 48: Why do we test all browser versions in fullbuildreduced?

Remotion/Build/Customizations/Projects.props
Stefan-Ilic: I gave it this name because these configurations are mainly there to ensure compatibility between different .NET execution environments, .NET target framework versions and SQL Server versions. I could remove the comments altogether, but to me, they were helpful

Build/TestingSetupForWebTest.targets
Stefan-Ilic: We'd have to ask @iam-mholle

Build/Shared.Server.Extensions.targets
Stefan-Ilic: FullBuildReduced is used for our git nightly builds, as I recall it was mainly there to run tests in all browsers in the first place

Remotion/Build/Customizations/Projects.props
MichaelKetting: new comment: exercise compatibility between installed .net version, target framework, and sql server

Build/TestingSetupForWebTest.targets
MichaelKetting: @Stefan-Ilic The build script has the option to select the docker image to use, correct? this information would then be used here. And of course, configured in the webtest-project configration so we test some browsers with .net 4.7.2 iis and some with .net 4.8 iis

Stefan-Ilic: The TestConfigurations have virtually nothing to do with the IIS hosting and they're basically only concerned with "Am I running nunit.exe inside a docker container or locally, and which TargetFrameworkVersion am I using for running said nunit.exe". In an instance of running a web test, no information regarding a docker image is available since web tests run locally.

Selecting the which docker image should be used for hosting IIS could certainly be added to TestConfigurations, but that would be a separate issue and is not possible ATM.

Off the top of my head, I can also think of solutions that could be done inside the trunk and wouldn't require adapting the BuildScript itself, but that should probably be its own issue as well.

MichaelKetting: Ah, I thought we already have the information as a property?
https://github.com/re-motion/SharedSource-Build/blob/develop/BuildScript/BuildTargets/Testing.targets#L90
` <_dockerImageName>%(_testOutputFiles.ExecutionRuntimeValue)</_dockerImageName>`

Can't we pipe this information to the `_dockerImageName` property in TestingSetupForWebTest.targets?

Stefan-Ilic: The problem is that if we're currently in a web test run, `_testOutputFiles.ExecutionRuntimeValue` is not a docker image but `LocalMachine` or `EnforcedLocalMachine`

Stefan-Ilic: The fact that it's still called `_dockerImageName` is probably just an oversight on my part

MichaelKetting: ah, i.c. Fell into the "web test configs are not ordered the same in terms of the target environment" descision. 

Okay, for the immediate future, we can use an ugly workaround by tying the docker image to the browser. that way, we still cover all .net versions in the fullbuild. 

am looking forward to learning to we can just drag the image name along into the test configuration. I know why and how we did things, but is it feasible to add the image name as an extra parameter? this would mean that "EnforcedLocalMachine" becomes a flag instead of an option next to "LocalMachine" and "Win_NET48". Then we could always expose the docker image if it's available (null/blank otherwise). This sounds like the cleanest approach to me but I may be biased and not considering all the options.

Stefan-Ilic: Tying the docker image and the browser together like `Chrome:Win_NET48` etc. was my first idea for a workaround as well, we'd probably be able to do that without touching the BuildScript and without causing more problems than we solve, but I agree that it wouldnt' be ideal.

We should probably get in a call tomorrow and hash this out, but I would still strongy suggest committing this as is right now so we can get to NUnit3 without falling into a black hole again and create a separate ticket for selecting the IIS hosting environment

MichaelKetting: I was actually thinking even simpler (more hackier) for line 35 in testingSetupForWebtests.targets
` <_dockerImageName Conditon=$(Browser) == "Edge">$(Docker_Win_Net48)</_dockerImageName>`
` <_dockerImageName Conditon=$(Browser) == "Chrome">$(Docker_Win_Net472)</_dockerImageName>`

Yes, it's an awful duplication but it's still only within the local build script and keeps things extremly simple and quick to merge. Design-review for the script extension task has time till next week.

Stefan-Ilic:
> 
> 
> Comments "Compatiblity Mode" are not done yet.

Yes, the browsers haven't been removed from SecurityManager either yet, buy thanks for reminding me! I'll re-request your review once it's done

Stefan-Ilic:
Added to `master` with d05b32323b18119821ce2dcae3327f70f88d8332

PR #73, opened 2020-04-22
RM-7412: Workaround for "NormalRequired_WithValidationErrors fails in IE with BocMultilineTextValue and BocTextValue"

Stefan-Ilic:
https://re-motion.atlassian.net/browse/RM-7412


Stefan-Ilic:
Added to `master` with 0f487fc97cabc4af9c656c85158fd931f058432c

PR #80, opened 2020-05-13
RM-7419 Remove period

dukorama:


Remotion/Web/Core/Globalization/DropDownMenuRenderer.fr.resx
```diff
@@ -121,6 +121,6 @@
     <value>Impossible d'afficher le menu.</value>
   </data>
   <data name="Remotion.Web.UI.Controls.DropDownMenuImplementation.Rendering.DropDownMenuRenderer.LoadingStatusMessage" xml:space="preserve">
-    <value>Charger.</value>
+    <value>Charger</value>
```
iam-mholle, line 124: Not related to this issue, but shouldn't this be "Chargement en cours" to match the other translations? Tell me my french experience pays off 😀 .

MichaelKetting: no clue. please compare with other single-word versions: Laden, Loading, and the Italian version. I'm assuming there is no need to do a three word version in French. You cannot compare the resource entry to the other French resource entries in this file without also cosnidering the source text in German and English.

iam-mholle: I compared the entry to the german ("Wird geladen") & italian ("Caricamento in corso") version of `LoadingStatusMessage` also affected in this PR. "Charger" would be "Laden".

MichaelKetting: ah, right. good catch. the single-word version is just in English. yeah, should be re-verified what the usual phrase is here.

dukorama: Already forwarded to Dejvid.

dukorama: You're right, Marcel, "Chargement en cours" would be correct.

PR #86, opened 2020-05-15
RM-7417: Nullable Reference Types in Core.Tools and Core.ServieLocation.PerformanceTestConsole

dkudernatsch:


Remotion/Core/Tools/Console/CommandLine/CommandLineArgument.cs
```diff
@@ -16,7 +16,7 @@
 // 
 using System;
 using System.Text;
-
+#nullable disable
```
MichaelKetting, line 19: why disable in procution code?

Remotion/Core/Tools/Console/CommandLine/CommandLineArgumentCollection.cs
```diff
@@ -16,7 +16,7 @@
 // 
 using System;
 using System.Collections;
-
+#nullable disable
```
MichaelKetting, line 19: why disable in procution code?

PR #87, opened 2020-05-15
RM-7417: Nullable Reference Types in Core.Reflection

dkudernatsch:


Remotion/Core/Reflection.CodeGeneration.UnitTests/ClassEmitterTest.cs
```diff
@@ -275,8 +275,8 @@ public void CreateEvent ()
     {
       var classEmitter = new CustomClassEmitter (Scope, "CreateEvent", typeof (object));
       CustomEventEmitter eventEmitter = classEmitter.CreateEvent ("Eve", EventKind.Instance, typeof (Func<string>));
-      eventEmitter.AddMethod.AddStatement (new ReturnStatement ());
-      eventEmitter.RemoveMethod.AddStatement (new ReturnStatement ());
+      eventEmitter.AddMethod?.AddStatement (new ReturnStatement ());
```
MichaelKetting, line 278: NFO here.

Remotion/Core/Reflection.CodeGeneration/CustomEventEmitter.cs
```diff
@@ -53,7 +53,7 @@ public CustomEventEmitter (CustomClassEmitter declaringType, string name, EventK
       declaringType.RegisterEventEmitter (this);
     }
 
-    public IMethodEmitter AddMethod
+    public IMethodEmitter? AddMethod
```
MichaelKetting, line 56: not null. please also re-check usages and remove nfo / ?

Remotion/Core/Reflection.CodeGeneration/CustomEventEmitter.cs
```diff
@@ -75,7 +75,7 @@ public IMethodEmitter AddMethod
       }
     }
 
-    public IMethodEmitter RemoveMethod
+    public IMethodEmitter? RemoveMethod
```
MichaelKetting, line 78: not null, see above.

Remotion/Core/Reflection.CodeGeneration/CustomPropertyEmitter.cs
```diff
@@ -70,7 +70,7 @@ public Type[] IndexParameters
       get { return _indexParameters; }
     }
 
-    public IMethodEmitter GetMethod
+    public IMethodEmitter? GetMethod
```
MichaelKetting, line 73: interesting. that's just...bizarre. not sure though, if it's also a null-bug in the design.

Remotion/Core/Reflection/NullMethodInformation.cs
```diff
@@ -28,22 +28,32 @@ namespace Remotion.Reflection
   /// </summary>
   public sealed class NullMethodInformation : IMethodInformation
   {
-    public string Name
+    public string? Name
```
MichaelKetting, line 31: The null-implementations need a revamp. The fact that the not-nullablity has been ignored is a bug.

Remotion/Core/Reflection/NullMethodInformation.cs
```diff
@@ -28,22 +28,32 @@ namespace Remotion.Reflection
...
+        return null;
+      }
+#nullable enable
     }
 
-    public ITypeInformation DeclaringType
+    public ITypeInformation? DeclaringType
     {
-      get { return null; }
+#nullable disable
```
MichaelKetting, line 43: why?

Remotion/Core/Reflection/NullMethodInformation.cs
```diff
@@ -28,22 +28,32 @@ namespace Remotion.Reflection
...
+      get 
+      {
+        return null;
+      }
+#nullable enable
     }
 
-    public ITypeInformation DeclaringType
+    public ITypeInformation? DeclaringType
```
MichaelKetting, line 41: DeclaringType is marked as nullable on the interface

Remotion/Core/Reflection/NullPropertyInformation.cs
```diff
@@ -18,6 +18,7 @@
 using System.Collections.Generic;
 using System.Reflection;
 
+#nullable disable
```
MichaelKetting, line 21: not disable

Remotion/Core/Reflection/NullPropertyInformation.cs
```diff
@@ -27,12 +28,18 @@ public sealed class NullPropertyInformation : IPropertyInformation
...
+      get 
+      {
+        return null;
+      }
     }
 
     public ITypeInformation DeclaringType
     {
-      get { return null; }
+      get
```
MichaelKetting, line 39: no reformat

Remotion/Core/Reflection/PropertyInfoAdapter.cs
```diff
@@ -163,16 +165,16 @@ public bool CanBeSetFromOutside
       return AttributeUtility.IsDefined<T> (_propertyInfo, inherited);
     }
 
-    public object GetValue (object instance, object[] indexParameters)
+    public object GetValue (object? instance, object[]? indexParameters)
     {
-      ArgumentUtility.CheckNotNull ("instance", instance);
+      ArgumentUtility.CheckNotNull ("instance", instance!);
```
MichaelKetting, line 170: null bug, see MethodInfoAdapter

Remotion/Core/Reflection/PropertyInfoAdapter.cs
```diff
@@ -163,16 +165,16 @@ public bool CanBeSetFromOutside
...
 
       return _propertyInfo.GetValue (instance, indexParameters);
     }
 
-    public void SetValue (object instance, object value, object[] indexParameters)
+    public void SetValue (object? instance, object? value, object[]? indexParameters)
     {
-      ArgumentUtility.CheckNotNull ("instance", instance);
+      ArgumentUtility.CheckNotNull ("instance", instance!);
```
MichaelKetting, line 177: null bug, see MethodInfoAdapter

Remotion/Core/Reflection/PropertyInfoAdapter.cs
```diff
@@ -232,7 +234,7 @@ public IEnumerable<IPropertyInformation> FindInterfaceDeclarations ()
       return _interfaceDeclarations.Value.ToArray();
     }
 
-    private IReadOnlyCollection<IPropertyInformation> FindInterfaceDeclarationsImplementation ()
+    private IReadOnlyCollection<IPropertyInformation?> FindInterfaceDeclarationsImplementation ()
```
MichaelKetting, line 237: not nullable

Remotion/Core/Reflection/PropertyInfoAdapter.cs
```diff
@@ -109,7 +111,7 @@ private PropertyInfoAdapter (PropertyInfo propertyInfo)
           LazyThreadSafetyMode.ExecutionAndPublication);
 
       _interfaceDeclarations = new Lazy<IReadOnlyCollection<IPropertyInformation>> (
-          FindInterfaceDeclarationsImplementation,
+          FindInterfaceDeclarationsImplementation!,
```
MichaelKetting, line 114: why?

Remotion/Core/Reflection/MethodInfoAdapter.cs
```diff
@@ -117,14 +119,14 @@ public ITypeInformation GetOriginalDeclaringType ()
       return AttributeUtility.GetCustomAttributes<T> (_methodInfo, inherited);
     }
 
-    public object Invoke (object instance, object[] parameters)
+    public object Invoke (object? instance, object?[]? parameters)
     {
-      ArgumentUtility.CheckNotNull ("instance", instance);
+      ArgumentUtility.CheckNotNull ("instance", instance!);
```
MichaelKetting, line 124: null bug
question is, do we want to add a conditional argument check or let the MethodInfo handle this? need test either way.

Remotion/Core/Reflection/ReflectionBasedMemberInformationNameResolver.cs
```diff
@@ -80,10 +81,12 @@ public string GetEnumName (Enum enumValue)
 
     private string GetPropertyNameInternal (IPropertyInformation propertyInformation)
     {
+#nullable disable
       return GetTypeName (propertyInformation.GetOriginalDeclaringType()) + "." + propertyInformation.Name;
+#nullable enable
     }
 
-    private string GetTypeNameInternal (ITypeInformation typeInformation)
+    private string? GetTypeNameInternal (ITypeInformation typeInformation)
```
MichaelKetting, line 89: That's a fun null bug :/

Remotion/Core/Reflection/NullMethodInformation.cs
dkudernatsch: What should i do with those, the interfaces declare something as not nullable and the implementation returns null. NFOing everything seems dangerous.

Remotion/Core/Reflection/PropertyInfoAdapter.cs
dkudernatsch: FindDeclaringProperty may return null per interface definition

Remotion/Core/Reflection/NullMethodInformation.cs
MichaelKetting: Right now, that's what we will do. The fact that we need to change this to something better is the bugfix-todo&jira-ticket

Remotion/Core/Reflection/PropertyInfoAdapter.cs
MichaelKetting: Hmm...I.C. 

Still wrong. If this were to occur, it would be a bug. It's not clear if the reflection implementation cannot trigger this scenario in the first place of if such members should be filtered. Fo now, we fix thsi with an NFO in the implementation and add a bugfix todo.

Remotion/Core/Reflection/MethodInfoAdapter.cs
```diff
@@ -117,14 +119,15 @@ public ITypeInformation GetOriginalDeclaringType ()
       return AttributeUtility.GetCustomAttributes<T> (_methodInfo, inherited);
     }
 
-    public object Invoke (object instance, object[] parameters)
+    public object Invoke (object? instance, object?[]? parameters)
     {
-      ArgumentUtility.CheckNotNull ("instance", instance);
+      //TODO RM-7432: Implementation requires not null on nullable parameter
```
MichaelKetting, line 124: something got lost somewhere in the comments, don't know where, don't know how, don't case.

here and in property info (GetValue, SetValue): the instance can be null, the null-check is wrong. -> otherwise, we don't support statics and that's just wrong.

Remotion/Core/Reflection/ITypeInformation.cs
```diff
@@ -249,7 +249,7 @@ public interface ITypeInformation : IMemberInformation
     /// or <see langword="null"/> if the current type is not an array or a pointer, or is not passed by reference, 
     /// or represents a generic type or a type parameter in the definition of a generic type or generic method.
     /// </returns>
-    [CanBeNull]ITypeInformation GetElementType ();
+    [CanBeNull] ITypeInformation? GetElementType ();
```
iam-mholle, line 252:
```suggestion
    [CanBeNull]ITypeInformation? GetElementType ();
```

Remotion/Core/Reflection/ReflectionBasedMemberInformationNameResolver.cs
```diff
@@ -67,7 +67,6 @@ public string GetPropertyName (IPropertyInformation propertyInformation)
     public string GetTypeName (ITypeInformation typeInformation)
     {
       ArgumentUtility.CheckNotNull ("typeInformation", typeInformation);
-
```
iam-mholle, line 70: Unnecessary change.

Remotion/Core/Reflection/ReflectionExtensions.cs
```diff
@@ -51,7 +51,7 @@ public static bool IsOriginalDeclaration (this IMemberInformation memberInfo)
     }
 
     [NotNull]
-    public static Type ConvertToRuntimeType (this ITypeInformation typeInformation)
+    public static Type? ConvertToRuntimeType (this ITypeInformation typeInformation)
```
iam-mholle, line 54: That's a bit contradictory 😉

Remotion/Core/Reflection.CodeGeneration/DPExtensions/ExpressionReference.cs
```diff
@@ -65,7 +65,7 @@ private LocalReference CreateLocal (ILGenerator gen)
       return local;
     }
 
-    public override void StoreReference (ILGenerator gen)
+    public override void StoreReference (ILGenerator? gen)
```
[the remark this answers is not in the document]
dkudernatsch: so, notnull and a nfo in tests?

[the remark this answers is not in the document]
MichaelKetting: yes. impl is not supported, value in test is probably a stub (I'm assuming) with ignored argument values. in generel: wenn the test passes dmmy values because IgnoreArguments, those are not indicative of the signatures nullablity.

Remotion/Core/Reflection.CodeGeneration/CustomPropertyEmitter.cs
MichaelKetting: not with the current implementation differences, we can't support a symmetrical API that's also safe. this difference might well be a bug and the get and set method be not-null. it's also possible that there's a subtle difference between event accessors and property accessors. either way, we can't repair this here and we can't fudge on the api without making it buggier. so, yes, prperty and event will be different until someone figures out if this is correct and fixes it in the correct order. given that Reflection.CodeGen is a legacy assembly, that won't happen anytime soon :D

PR #91, opened 2020-05-25
RM-7431: Update to C#8 MSBuild 2019

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7431


Remotion/Core/Collections.Caching/Core.Collections.Caching.csproj
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="15.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="Current" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
MichaelKetting, line 2: Okay, microsoft did a big doodoo with the toolsversion. Before, I was using the toolsversion to use a specific msbuild tools version when building. with msbuild 2019 onward this is no longer possible
(https://developercommunity.visualstudio.com/content/problem/404485/vs2019-msbuildtoolsversion-is-not-a-version.html)

Using the correct term to describe our situation is not prudent within a public forum, I will not.

https://docs.microsoft.com/en-us/visualstudio/msbuild/whats-new-msbuild-16-0?view=vs-2019

In build.targets, we extract the toolsversion from the project file and assign it to the msbuild call. this was done to ensure we're usign the same msbuiild tools version even when upgrading the msbuild "runtime" used to execute the build scripts. we will need to follow this up on the msbuild github issue tracker and ask how the tools version can be used for backwards compat in the future (and how this will work with sdk style projects :/ )

MichaelKetting: https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-toolset-toolsversion?view=vs-2019

Apparently, once we switch to sdk style, this "might" go away?

MichaelKetting: https://re-motion.atlassian.net/browse/RMSRCBUILD-126

iam-mholle: @MichaelKetting despite all the issues, this is ready to merge right? I didn't quite understand why/when we would need the ability to specify an older `ToolsVersion`..

MichaelKetting: Yes, we can merge this. The problem with the toolsversion is, that we can't do builds that target msbuild 2019 once they release msbuild 2021 or whatever when using said msbuild 2021 as the "root" executable for running the build. This means, when we update the tooling in the templates to msbuild 2021 and we don't have a good alternative setup usign the sdk parameter or something of the sort, upgrading the toling may introduce changes to the c# compiler which in turn may change the behavior of the support branch, causing unforseen bugs. And yes, c# already introduced really nasty bugs in the past.

iam-mholle:
Added to master with c035603dcd40f618222914c8467a452e020f061d.

PR #75, opened 2020-05-06
RM-7417: Nullable Reference Types in Core.Core

dkudernatsch:
https://re-motion.atlassian.net/browse/RM-7417

PR #92, opened 2020-05-25
RM-7435: Marked StringUtility.ConcatWithSeparator overload obsolete

dkudernatsch:
https://re-motion.atlassian.net/browse/RM-7435

### DIFF
--- a/Remotion/Core/Core/Utilities/StringUtility.cs
+++ b/Remotion/Core/Core/Utilities/StringUtility.cs
@@ -275,9 +275,12 @@ namespace Remotion.Utilities
 
     public static string ConcatWithSeparator (IList list, string separator)
     {
+#pragma warning disable 618
       return ConcatWithSeparator (list, separator, null, null);
+#pragma warning restore 618
     }
 
+    [Obsolete ("Use ConcatWithSeperator (IList, string) instead. Parameter 'format' is no longer used.")]
     public static string ConcatWithSeparator (IList list, string separator, string format, IFormatProvider formatProvider)
     {
       if (list == null)
@@ -343,7 +346,9 @@ namespace Remotion.Utilities
       string format = null;
       if (elementType == typeof (float) || elementType == typeof (double))
         format = "R";
+#pragma warning disable 618
       return ConcatWithSeparator ((IList) value, ",", format, formatProvider);
+#pragma warning restore 618
     }
 
     private static string FormatScalarValue (object value, IFormatProvider formatProvider)
--- a/Remotion/Core/Core/Utilities/StringUtility.cs
+++ b/Remotion/Core/Core/Utilities/StringUtility.cs
@@ -280,7 +280,7 @@ namespace Remotion.Utilities
 #pragma warning restore 618
     }
 
-    [Obsolete ("Use ConcatWithSeperator (IList, string) instead. Parameter 'format' is no longer used.")]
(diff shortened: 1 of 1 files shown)


Remotion/Core/Core/Utilities/StringUtility.cs
```diff
@@ -275,9 +275,12 @@ public static string[] ListToStringArray (IList list)
 
     public static string ConcatWithSeparator (IList list, string separator)
     {
+#pragma warning disable 618
       return ConcatWithSeparator (list, separator, null, null);
+#pragma warning restore 618
     }
 
+    [Obsolete ("Use ConcatWithSeperator (IList, string) instead. Parameter 'format' is no longer used.")]
```
iam-mholle, line 283: The version introducing the obsolete attribute needs to be added after the message.

```suggestion
    [Obsolete ("Use ConcatWithSeperator (IList, string) instead. Parameter 'format' is no longer used. (Version 1.21.8)")]
```

Remotion/Core/Core/Utilities/StringUtility.cs
```diff
@@ -275,9 +275,12 @@ public static string[] ListToStringArray (IList list)
 
     public static string ConcatWithSeparator (IList list, string separator)
     {
+#pragma warning disable 618
       return ConcatWithSeparator (list, separator, null, null);
+#pragma warning restore 618
     }
 
+    [Obsolete ("Use ConcatWithSeperator (IList, string) instead. Parameter 'format' is no longer used.")]
     public static string ConcatWithSeparator (IList list, string separator, string format, IFormatProvider formatProvider)
```
iam-mholle, line 284: Imo another overload should be added, which has the `format` parameter removed.
`public static string ConcatWithSeparator (IList list, string separator, IFormatProvider formatProvider)`

The 2 calls of the old method enclosed in `pragma` as well as the obsolete method should then call the method with the new signature. Or is this another issue? @MichaelKetting, quoting https://github.com/re-motion/Trunk-SVN2Git/pull/75/files#r428597852

MichaelKetting: since there's no one using this, this helper should just be removed. hence the obsolete message for now. Since there's a ton of no longer needed legacy stuff on StringUtility, a clean-up should be a coordinated task, not just a single method.

iam-mholle: Got it

Remotion/Core/Core/Utilities/StringUtility.cs
```diff
@@ -535,4 +540,4 @@ public ParseException (SerializationInfo info, StreamingContext context)
     {
     }
   }
-}
\ No newline at end of file
+}
```
iam-mholle, line 543: This happens when doing a suggestion commit via Github, dunno if we keep this.

MichaelKetting: don't care

Remotion/Core/Core/Utilities/StringUtility.cs
iam-mholle: Added https://re-motion.atlassian.net/browse/RM-7436

iam-mholle:
Added to master with bc5e405d773d91e0b5d028d39943e0795f1e2ffd.

PR #94, opened 2020-05-29
RM-3427: Use DateTime.UtcNow instead of DateTime.Now for function timeouts

iam-mholle:
https://re-motion.atlassian.net/browse/RM-3427


iam-mholle:
Added to master with 8fb70f4fc1f98631ba520a28da5c3d752ae33d35.

PR #95, opened 2020-05-29
RM-7417: Fixed compile error due to new msbuild version

dkudernatsch:


iam-mholle:
@MichaelKetting what's the plan here - this is possibly a roslyn bug. NFO + TODO comment for later improvements?

Signature in NET Core for reference: `bool Equals([AllowNull] T other);` -> so apart from being an invocation of a null oblivious method, this should be legal without NFO anyway, in Core at least.

MichaelKetting:
@iam-mholle Sounds sensible with the NFO. Please follow up on this with a Roslyn ticket, a cleanup-ticket in our jira and the appropriate todo comment. Oh, and we need to verify this against the official .net 5 asemblies, to make sure this isn't somehting we mucked up with the reference assemblies.

iam-mholle:
@MichaelKetting Nevermind. This is the expected behavior.

> [CS8604] Possible null reference argument for parameter 'other' in 'bool IEquatable<T>.Equals(T other)'.

`T` is unconstrained nullability-wise, but constrained to `where T : IEquatable<T>`. The parameters are marked to `[AllowNull]`, so the following call could happen: `IEquatable<T!>.Equals((T!) null)`, that's why we get the error, nothing to do with null oblivious code wrongly analyzed.

PR #97, opened 2020-06-04
RM-7441: VerifyPEFile returns Error 0x80070002 "The system cannot find the file specified"

Stefan-Ilic:
https://re-motion.atlassian.net/browse/RM-7441

### DIFF
--- a/Remotion/Mixins/UnitTests/Core/CodeGeneration/SetUpFixture.cs
+++ b/Remotion/Mixins/UnitTests/Core/CodeGeneration/SetUpFixture.cs
@@ -71,9 +71,8 @@ namespace Remotion.Mixins.UnitTests.Core.CodeGeneration
     [OneTimeSetUp]
     public void OneTimeSetUp ()
     {
-      Environment.CurrentDirectory = TestContext.CurrentContext.TestDirectory;
       var assemblyTrackingPipelineFactory = new AssemblyTrackingPipelineFactory();
-      var settings = PipelineSettings.New().SetEnableSerializationWithoutAssemblySaving (true).Build();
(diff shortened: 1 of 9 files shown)


Remotion/Mixins/UnitTests/Core/CodeGeneration/SetUpFixture.cs
```diff
@@ -71,6 +71,7 @@ public static void AddSavedAssembly (string assemblyPath)
     [OneTimeSetUp]
     public void OneTimeSetUp ()
     {
+      Environment.CurrentDirectory = TestContext.CurrentContext.TestDirectory;
```
MichaelKetting, line 74: we have been able to work (almost) compeltely without setting Env.CurDir when @dkudernatsch did the other conversions. we should try to keep it this way.

See https://re-motion.atlassian.net/browse/RM-7445

Please check if it is really not sensibly possible to use absolute paths in this instance. If not, please add to the aformentioned jira ticket.

Stefan-Ilic: I'll take another look at this next week, but some `Assembly.Save` method (if I remember correctly) only accepts filenames and no paths, which makes it a little harder

iam-mholle: I thought about the same thing while reviewing, but could not find a nice way to work around setting `Environment.CurrentDirectory`.

MichaelKetting: Should be timeboxed.

MichaelKetting: okay, I think this might be a one-liner. Didn't try it yet, but I know what to look for in the next review :)

Stefan-Ilic:
Added to `master` with 4476efd27e5a7b07590491631152316924137b5d

PR #107, opened 2020-06-26
RM-7444: Inline CheckNotNull call

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7444


iam-mholle:
Added to master with 020ba6d0de4bbb347975aaa2a118fcefd631cbfa.

PR #109, opened 2020-07-09
RM-7458: Return original hostname if the loopback address is resolved

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7458


iam-mholle:
Added to master with 5dcaa8c5de659a359e6942f142eab688c14f3b13.

PR #110, opened 2020-07-14
RM-7446: Right textbox border cutoff since Chrome v83

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7446

Reference screenshots are not yet updated.


MichaelKetting:
Regarding the question if we should use width:calc-2px or add a padding, I'll need to see this live. Based on the existing css, we already use calc for calculating widths/heights, so it's jsut a question if padding or width is the correct / less-intrustive option here.

Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocAutoCompleteReferenceValue.css
```diff
@@ -24,7 +24,8 @@ span.bocAutoCompleteReferenceValue input
 /* Sets width for Chromium browsers */
```
MichaelKetting, line 24: .blink already says that this is for chromium. If we want a comment, we should document what the workaround fixes.

Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocDateTimeValue.css
```diff
@@ -202,3 +182,11 @@ span.bocTimeInputWrapper input
   height: 1.2em;
   line-height: 1em;
 }
+
+/* Sets width for Chromium browsers */
```
MichaelKetting, line 186: see above.

Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocAutoCompleteReferenceValue.css
```diff
@@ -24,7 +24,8 @@ span.bocAutoCompleteReferenceValue input
 /* Sets width for Chromium browsers */
 .blink span.bocAutoCompleteReferenceValue input
```
MichaelKetting, line 25: should we move this to the bottom of the file, similiar to the other controls?

Remotion/ObjectBinding/Web/Res/Themes/NovaGray/Html/BocTextValue.css
```diff
@@ -70,7 +70,8 @@ span.bocTextValue input
 /* Sets width for Chromium browsers */
```
MichaelKetting, line 70: see above

MichaelKetting:
please also add a generic input[type=text] and type=password to style.css. I know it's redundant, but this will catch all the other textboxes that might exist.

iam-mholle:
Added to master in d67113c3abdeabb191152fcad044789d5c4cad77.

PR #115, opened 2020-08-07
RM-7467: ProcessSnapshot filter

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7467


MichaelKetting:
Since this also fixes the root for rm-7420, should we revert the retry-loops?

iam-mholle:
> 
> 
> Since this also fixes the root for rm-7420, should we revert the retry-loops?
> 
> Edit: okay, I think that's already done.

That's part of the PR.

https://github.com/re-motion/Trunk-SVN2Git/pull/115/files#diff-916fd6f6bc7a90895d1da479ef9019b5L34

iam-mholle:
> 
> 
> Do we have production code, too, that relies on finding processes/ comparing them, waiting for them to exit?

Did not find any. `Remotion.Web.Development.WebTesting.Utilities.ProcessUtils.KillAllProcessesWithName` is using `Process.GetProcessesByName(string)` only for killing them afterwards.

iam-mholle:
Added to master with 532b4a50761bc415f091c7ce5e0845747dcb1a74.

PR #113, opened 2020-08-04
RM-7464 (fix): Rework BrowserSessionBase.Dispose

iam-mholle:
Browser subprocesses were sometimes not shutdown in time, there was no kill issued before this PR. The usage of `ProcessUtils.GracefulProcessShutdown` ensures the process shutdown.

This most likely also fixes the flaky tests `WebTestHelper_OnTestFixtureTearDown_CleansUserDirectory_Chromium(True)` & `WebTestHelper_OnTestFixtureTearDown_CleansUserDirectory_Chromium(False)`, where a process still had a user file locked.

PR #116, opened 2020-08-07
RM-7468: ProcessUtils.GracefulProcessShutdown does not handle Win32Exception when killing a terminating process

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7468

### DIFF
--- a/Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
+++ b/Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
@@ -138,6 +138,7 @@ namespace Remotion.Web.Development.WebTesting.Utilities
       // Try to gracefully close the process
       if (process.CloseMainWindow())
       {
+        // No exception is thrown when calling .WaitForExit(..) on an already exited process.
         if (process.WaitForExit (timeoutInMilliseconds))
           return;
       }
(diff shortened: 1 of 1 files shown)


Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
```diff
@@ -143,7 +143,16 @@ public static void GracefulProcessShutdown ([NotNull] Process process, TimeSpan
...
+        process.Kill();
+      }
+      catch (Win32Exception)
+      {
+        // Thrown if the .Kill() method is called while the process is currently terminating.
+        // https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.kill#remarks
+        // This exception can be swallowed safely, as the process that should be killed is already terminating.
+      }
 
       if (process.WaitForExit (timeoutInMilliseconds))
```
MichaelKetting, line 157: Hmm...I think if the process is already dead at this point, we might get another exception (see docs). this one should be easy to test with a debugger to delay the call to WaitForExit until the process is gone. don't know if the currently being-killed process will also throw, same as in line 148, though, or if it's just a missing process that will throw. =-> check docs, do trials, make safe.

iam-mholle: That's an issue we must have met before the removal of this line: https://github.com/re-motion/Trunk-SVN2Git/pull/113/files#diff-3569e8db6d4a2bdf25ec2682ebd29c24L127

Pretty sure this is a non-issue.

iam-mholle: Yep, tested while asserting that the process has exited.

Remotion/Web/Development.WebTesting/Utilities/ProcessUtils.cs
```diff
@@ -143,7 +143,16 @@ public static void GracefulProcessShutdown ([NotNull] Process process, TimeSpan
       }
 
       // Force closing the process
-      process.Kill();
+      try
+      {
+        process.Kill();
+      }
+      catch (Win32Exception)
```
MichaelKetting, line 150: One more:
https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.kill?view=netframework-4.8

Kill is said to throw an InvalidOpEx if the process has alreayd existed. Please verify.

MichaelKetting: Oh, and we should probably doc all the non-issue exceptions regardless if they are mentioned in the docs but do not apply for non-obvious reasons.

iam-mholle: That's true, I'll add another `catch` block.

iam-mholle:
Added to master with 3e648b9240a3a00c6d87f25194a2be59ae6820bc.

PR #122, opened 2020-09-02
RM-7476: Suppress double postbacks not triggered via autopostback + not autopostback

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7476

### DIFF
--- a/Remotion/Web/Core/Res/HTML/SmartPage.js
+++ b/Remotion/Web/Core/Res/HTML/SmartPage.js
@@ -1176,7 +1176,7 @@ function SmartPage_Context(
     };
     if (_submitState == null)
       _submitState = submitState;
-    else if (isAutoPostback || submitterElement == null || submitterElement.id !== _submitState.Submitter.id)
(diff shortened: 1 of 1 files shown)


Remotion/Web/Core/Res/HTML/SmartPage.js
```diff
@@ -1176,7 +1176,7 @@ function SmartPage_Context(
     };
     if (_submitState == null)
       _submitState = submitState;
-    else if (isAutoPostback || submitterElement == null || submitterElement.id !== _submitState.Submitter.id)
+    else if (_submitState.IsAutoPostback && !isAutoPostback)
```
Stefan-Ilic, line 1179: Bump for continuity: when filling in a textbox, then hitting the async link and then hitting the async button, it does a postback for the textbox and the button, but not for the link (which i clicked before the button). Please report back to Michael to check if that behavior is intended.

MichaelKetting (approved):

iam-mholle:
Added to master with d039561dfd2835dec9a081536c23dae3a5c7efd7.

PR #123, opened 2020-09-07
RM-6490: Fix AutoCompleteReferenceValue postback queuing

iam-mholle:
https://re-motion.atlassian.net/browse/RM-6490

Tested the scenarios described in the issue:
1. When the value is cleared from the Textbox with IE's "reset" control ("x"), and the button is clicked, the Textbox's AutoPostBack is executed, then the button's postback is executed.
2. When the value is cleared from the Textbox manually, and the button is clicked, the Textbox's AutoPostBack is executed, then the button's postback is executed.
3. When the value is cleared in the currently focused BocAutoCompleteReferenceValue by selecting the text using the mouse and then pressing Backspace or Delete before clicking the button, the TextBox is posted as an empty value and the hidden field holds the value `==null==`.

### DIFF
--- a/Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js
+++ b/Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js
@@ -360,10 +360,8 @@
                 } else {
                     clearTimeout(timeout);
                     var lastKeyPressCode = state.lastKeyPressCode;
-                    timeout = setTimeout(function () {
-                        invalidateResult();
-                        acceptInput(lastKeyPressCode);
-                    }, 200);
+                    invalidateResult();
+                    acceptInput(lastKeyPressCode);
                 }
             }
         }).click(function() {


Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js
```diff
@@ -229,7 +229,6 @@
                 case KEY.ESC:
                     // re-motion: block event bubbling
                     event.stopPropagation();
-                    clearTimeout(timeout);
```
iam-mholle, line 232: This is covered by 
https://github.com/re-motion/Trunk-SVN2Git/blob/8dd67a6ad3a3b7e38bb6f528eb1e971fef64ad9e/Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js#L168

I could not find any scenario where this line would be needed.

Stefan-Ilic (approved):
I tested all 3 scenarios described in the issue with all browsers and it all seems to work fine. Nice job!

iam-mholle:
Added to master with 9c479a9a0818f7d0abfd743b572bac70d6310a78.

MichaelKetting:
Reverted on master with 45bbe1b33a5e756eade048f9c2a3914e4214bcf1

Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js
MichaelKetting: Hmm.... I think it should be a separate issue (task), just to document this stuff. Otherwise, autocomplete will kill us with special case research.

MichaelKetting (changes requested):

MichaelKetting (approved):

iam-mholle:
Added to master with f8491379c772c142d63aebafc184c2208d976818.

PR #124, opened 2020-09-10
RM-7494: The second postback of a double postbacks seems to be preferrably a button rather than a link

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7494


MichaelKetting (changes requested):
As discussed, OnFormClick is called for every click on the aera covered by the form. In addition, CheckFormState() calls Form.OnSubmit, this may result in side effects if called multiple times.

iam-mholle:
Closed in favor of #148.

PR #125, opened 2020-09-14
RM-7491: Upgrade Firefox to Version 80

Stefan-Ilic:
https://re-motion.atlassian.net/browse/RM-7491

Breaking Changes: Download of zero-size files is not supported anymore.


Remotion/Web/Development.WebTesting.IntegrationTests/ScreenshotTest.cs
```diff
@@ -406,7 +406,7 @@ public void DropDownList ()
       }
       else if (Helper.BrowserConfiguration.IsFirefox())
       {
-        padding = new WebPadding (2, 1, 2, 68);
+        padding = new WebPadding (6, 1, 2, 68);
```
iam-mholle, line 409: The width of an empty dropdown changed.

image

iam-mholle (approved):

Remotion/Web/Development.WebTesting/DownloadInfrastructure/DownloadedFileFinder.cs
```diff
@@ -247,5 +222,23 @@ private string GetPartialFile (IEnumerable<string> newFiles)
     {
       return newFiles.SingleOrDefault (file => file.EndsWith (_partialFileExtension));
     }
+
+    private bool PartialFileExists (IEnumerable<string> newFiles)
+    {
+      return GetPartialFile (newFiles) != null;
+    }
+
+    private bool PartialFileWasFoundBefore ()
```
iam-mholle, line 231: The name `PartialFileWasFoundBefore` is not ideal imo, got no better suggestion tho. I guess adding docs here would help.

iam-mholle (changes requested):

Remotion/Web/Development.WebTesting/DownloadInfrastructure/DownloadedFileFinder.cs
Stefan-Ilic: I would really like to avoid additional comments. I've renamed the method again, though. Take a look and tell me what you think

iam-mholle (approved):

MichaelKetting:
> 
> 
> https://re-motion.atlassian.net/browse/RM-7491
> 
> Breaking Changes: Download of zero-size files is not supported anymore.

@Stefan-Ilic Please add Breaking Change doc. Sometimes (depends on the timeframe) I add this information inline into the description of the jira ticket and then make a real breaking change ticket from this and remove the line once more. Adding it as a note in GitHub is okay, too, provided it's not lost in the discussion thread. Maybe do it as  a task-list? Or update the PR description itself?

Remotion/Web/Development.WebTesting/DownloadInfrastructure/DownloadedFileFinder.cs
```diff
@@ -242,10 +222,29 @@ private bool FileGotUpdated (string oldPartialFile, DateTime oldPartialFileLastW
...
-      return newFiles.SingleOrDefault (file => file.EndsWith (_partialFileExtension));
+      return _partialFileStateOfCurrentDownload != null;
+    }
+
+    private bool IsZeroLengthFile (string fileName)
+    {
+      var filePath = Path.Combine (_downloadDirectory, fileName);
+      var fileInfo = new FileInfo (filePath);
+
+      return  fileInfo.Length == 0;
```
MichaelKetting, line 247: double whitespace

Remotion/Web/Development.WebTesting/DownloadInfrastructure/DownloadedFileFinder.cs
```diff
@@ -242,10 +222,29 @@ private bool FileGotUpdated (string oldPartialFile, DateTime oldPartialFileLastW
...
-    [CanBeNull]
-    private string GetPartialFile (IEnumerable<string> newFiles)
+    private bool TryGetPartialFile (IEnumerable<string> newFiles, out string partialFile)
+    {
+      partialFile = newFiles.SingleOrDefault (file => file.EndsWith (_partialFileExtension));
+
+      return partialFile != null;
+    }
+
+    private bool PartialFileExists (IEnumerable<string> newFiles)
```
MichaelKetting, line 232: HasPartialFile

Remotion/Web/Development.WebTesting/DownloadInfrastructure/DownloadedFileFinder.cs
```diff
@@ -108,71 +107,45 @@ public FileInformationTuple (DateTime lastWriteTimeUtc, long length)
       {
         var newFiles = GetNewFiles (filesInDownloadDirectoryBeforeDownload, startedDownloadHandlingUtc);
 
-         //The first step, we haven't found a partial File yet to monitor the download process
-        if (partialFileState == null)
+        if (TryGetPartialFile (newFiles, out var partialFile)
```
MichaelKetting, line 110: Some of the API choices on PartialFileState would make for a great refactoring exercise. Maybe this could be a Mob-session sometime?

Remotion/Web/Development.WebTesting/DownloadInfrastructure/DownloadedFileFinder.cs
```diff
@@ -55,7 +55,7 @@ public FileInformationTuple (DateTime lastWriteTimeUtc, long length)
     private readonly string _partialFileExtension;
     private readonly IDownloadFileFinderStrategy _downloadFileFinderStrategy;
     private readonly TimeSpan _downloadStartedGracePeriod;
-
+    private PartialFileState _partialFileStateOfCurrentDownload;
```
MichaelKetting, line 58: Looking at the final code, I don't see a need to have this as a field. We can always pass the value as a parameter (there's just two methods that need this). 

Reasoning: 
* Before this change, DownloadedFileFinder does not have mutable state. It doesn't matter if we create a new one for each request or if there's no reasonable scenario where you would want to run two of them in parallel. It still keeps the state and the behavior of the type pure if we don't introduce mutable state. 
* When we do one day refactor this, the cut line will be along the state object (since it's already aptly named) and mutable state must then gravitate towards this type. Having the state as variables instead of fields on another type, keeps the state isoated.
* It feels like it's the right thing to do (TM).

Remotion/Web/Development.WebTesting/DownloadInfrastructure/DownloadedFileFinder.cs
```diff
@@ -108,71 +107,45 @@ public FileInformationTuple (DateTime lastWriteTimeUtc, long length)
...
+                 && PartialFileWasFoundInPreviousIteration()
+                 && PartialFileWasUpdated()
+                 && TryGetFileInformation (_partialFileStateOfCurrentDownload.GetPartialFile(), out var fileInfoOfPartialFile))
+        {
+          _partialFileStateOfCurrentDownload.UpdatePartialFileLastWriteAccessUtc (fileInfoOfPartialFile.LastWriteTimeUtc);
+          _partialFileStateOfCurrentDownload.UpdatePartialFileLength (fileInfoOfPartialFile.Length);
+
+          downloadTimeWithoutUpdate.Restart();
+        }
+        else if (!PartialFileExists (newFiles) && !TemporaryFilesExist (newFiles) && newFiles.Count > 0 && newFiles.Any (IsZeroLengthFile))
```
MichaelKetting, line 130: Isn't this equivalent to writing newFiles.Any(IsZeroLengthFile) ?

Reasoning: if any of the other condiitons is false, it will fall through the next if-else, too. 

It may still be a relevant bit of documentation, but that may be true, or not.

MichaelKetting (changes requested):
I explicitly did not try to reason (much) on the download-detection logic. I am assuming that the tests will tell us if it works.

Remotion/Web/Development.WebTesting/DownloadInfrastructure/DownloadedFileFinder.cs
Stefan-Ilic: Would the starting point be the old design or the "simplified" version in this PR?

MichaelKetting: I think we can start from here. The mainly because it looks like the refactoring helped and I don't want to spend a week in a mob training session.

Remotion/Web/Development.WebTesting/DownloadInfrastructure/DownloadedFileFinder.cs
Stefan-Ilic: I tried thinking about this but the reality is that without the other conditions, the tests fail

MichaelKetting: That's...strange. something for the mob session, then.

Stefan-Ilic: Yeah, this entire PR for me was "Let me make this code cleaner, oh wait, it doesn't work anymore" so I'd suggest I present you with a solution that is *stable* and *readable* and we take mental notes of the *optimal* way for the mob session 😄

MichaelKetting: okay, the new code is different for the two branch conditions, so the original assumption no longer holds.

MichaelKetting (approved):
please wait with the merge to svn until tomorrow so we can do the fullbuilds without this. just to make sure there's no new moving parts

Stefan-Ilic:
> 
> 
> please wait with the merge to svn until tomorrow so we can do the fullbuilds without this. just to make sure there's no new moving parts

This solution is stable locally, but fails the 'ol "run every download test 1k times routine" So I still have to run regular Firefox builds a reasonable number of times before I can even be sure enough that it's stable in the long run

Stefan-Ilic:
Added to `master` with 6259792ef57c99f0f24b3677538cd6aad7f72640

PR #127, opened 2020-09-18
RM-7503: Move DropDownMenu options next to the DropDownMenu button instead of the bottom of the body

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7503


dukorama (approved):

Remotion/Web/Core/Res/HTML/DropDownMenu.js
```diff
@@ -286,7 +286,7 @@ function DropDownMenu_BeginOpenPopUp(menuID, context, evt)
     statusPopup.setAttribute('aria-labelledby', menuButton[0].id);
   _dropDownMenu_currentStatusPopup = statusPopup;
   $(statusPopup).iFrameShim({ top: '0px', left: '0px', width: '100%', height: '100%' });
-  document.body.appendChild(statusPopup);
+  document.getElementById(menuID).parentElement.appendChild(statusPopup);
```
MichaelKetting, line 289: problem: context-menu for TreeView (e.g. SingleTestTreeView.aspx) also opens a popup menu. Parent element is a span, popup is a div, this this invalid HTML.

MichaelKetting (changes requested):

Remotion/Web/Core/Res/HTML/DropDownMenu.js
iam-mholle: True. I guess the dropdown menu does not need to be a `div` anyway right? So that would be one way to fix it. I could also walk up the DOM until an `UpdatePanel` is found (or the end is reached) and place the popup there.

MichaelKetting: The dropdown element must be a div. There's an UL inside, so we have to nest it in a blocklevel element.

Correct about the walk. 

Two options:
a) walk up until you find a DIV or TD element (or root) and be done with it.

b) Update Panel IDs are available on async-Pages via {{Sys.WebForms.PageRequestManager.getInstance()._updatePanelIDs}}, which is a tad hacky since the _updatePanelIDs is not actually a public API. It probably won't matter, though. The problem with this approach though, is that the popup is a blocklevel-element, so if the updatepanel is a SPAN, it won't be correct html anyway. This means the behavior would be correct, but the html markup won't be. In theory, we could add a logstatement if we find this scenario and just print an error if the popup doesn't properly work with the currect page structure.

iam-mholle: Well a) sounds better imo. Will do.

MichaelKetting: DIV / TD / TH

iam-mholle:
Note to myself: `.closest(...)` does not work in IE.

MichaelKetting (approved):

iam-mholle:
Added to master with 0a34fc3743cde760566148a9cf491e7d015a0433.

MichaelKetting:
Reverted on master with 2c73133d30d854707703f9470a1e10c1943e46a6.

Remotion/Web/Development.WebTesting.IntegrationTests/Web.Development.WebTesting.IntegrationTests.csproj
```diff
@@ -173,8 +173,7 @@
     <EmbeddedResource Include="SavedTestScreenshots\DropDownListControlObjectTest\InternetExplorer.any.DropDownList0.png" />
     <EmbeddedResource Include="SavedTestScreenshots\DropDownListControlObjectTest\InternetExplorer.any.DropDownList1.png" />
     <EmbeddedResource Include="SavedTestScreenshots\DropDownMenuControlObjectTest\DropDownMenu.png" />
-    <EmbeddedResource Include="SavedTestScreenshots\DropDownMenuControlObjectTest\InternetExplorer.any.DropDownMenu0.png" />
-    <EmbeddedResource Include="SavedTestScreenshots\DropDownMenuControlObjectTest\InternetExplorer.any.DropDownMenu1.png" />
+    <EmbeddedResource Include="SavedTestScreenshots\DropDownMenuControlObjectTest\InternetExplorer.any.DropDownMenu.png" />
```
MichaelKetting, line 176: I'm confused about 
Remotion/Web/Development.WebTesting.IntegrationTests/SavedTestScreenshots/DropDownMenuControlObjectTest/DropDownMenu.png
and
Remotion/Web/Development.WebTesting.IntegrationTests/SavedTestScreenshots/DropDownMenuControlObjectTest/Firefox.any.DropDownMenu.png

Those have been deleted but no new files added? the test failed for all browsers, so how does this work? also, if they've been deleted, where's the change in the csproj file?

MichaelKetting (changes requested):

Remotion/Web/Development.WebTesting.IntegrationTests/Web.Development.WebTesting.IntegrationTests.csproj
iam-mholle: They have both been updated, not deleted. Try enabling the rich diff for those files.

MichaelKetting: rich diff, does this exist on github review?

MichaelKetting (approved):

Remotion/Web/Development.WebTesting.IntegrationTests/Web.Development.WebTesting.IntegrationTests.csproj
iam-mholle: image

MichaelKetting: That's sooooo obvious...

iam-mholle:
Added to master with c8239773f6689d8649b6ff3f6df21e8d09417e55.

PR #129, opened 2020-09-24
RM-7485: Upgrade Edge to Version 85

Stefan-Ilic:
Raffael, please don't peek, we'll go through the changes like I never spiked them in the first place

https://re-motion.atlassian.net/browse/RM-7485

### DIFF
--- a/Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/EdgeBrowserContentLocator.cs
+++ b/Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/EdgeBrowserContentLocator.cs
@@ -131,12 +131,14 @@ namespace Remotion.Web.Development.WebTesting.ScreenshotCreation.BrowserContentL
 
     private AutomationElement GetContentElement (AutomationElement window)
     {
-      return window.FindFirst (
-          TreeScope.Subtree,
-          new AndCondition (
-              new PropertyCondition (AutomationElement.ControlTypeProperty, ControlType.Pane),
-              new PropertyCondition (AutomationElement.ClassNameProperty, "View"),
-              new PropertyCondition (AutomationElement.FrameworkIdProperty, c_edgeFrameworkID)));
+      return window.FindAll (
+              TreeScope.Subtree,
+              new AndCondition (
+                  new PropertyCondition (AutomationElement.ControlTypeProperty, ControlType.Pane),
+                  new PropertyCondition (AutomationElement.ClassNameProperty, "View"),
+                  new PropertyCondition (AutomationElement.FrameworkIdProperty, c_edgeFrameworkID)))
+          .Cast<AutomationElement>()
+          .Aggregate (GetElementWithLargerArea);
     }
 
     private TResult RetryUntilValueChanges<TResult> (Func<TResult> func, TResult value, int retries, TimeSpan interval)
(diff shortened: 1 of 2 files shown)


Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/CustomWebTestConfigurationFactory.cs
```diff
@@ -56,7 +56,7 @@ protected override IEdgeConfiguration CreateEdgeConfiguration (WebTestConfigurat
       if (string.IsNullOrEmpty (edgeVersionArchivePath))
         return new EdgeConfiguration (configSettings);
 
-      var versionedEdgeFolder = $"Edge_v{LatestTestedEdgeVersion}";
+      var versionedEdgeFolder = $"Edge_v{LatestTestedEdgeVersion}_1";
```
Stefan-Ilic, line 59: Development change, need to revert before merge

iam-mholle: Don't forget to do this.

iam-mholle (approved):
💯

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/EdgeBrowserContentLocator.cs
```diff
@@ -153,5 +155,21 @@ private TResult RetryUntilValueChanges<TResult> (Func<TResult> func, TResult val
 
       return value;
     }
+
+    private AutomationElement GetElementWithLargerArea (AutomationElement firstAutomationElement, AutomationElement secondAutomationElement)
+    {
+      var firstRectangle = firstAutomationElement.Current.BoundingRectangle;
+      var secondRectangle = secondAutomationElement.Current.BoundingRectangle;
+
+      if (firstRectangle == Rect.Empty)
```
MichaelKetting, line 164: aren't those two redundant and actually introduce more complexity given they are covered by the condition in line 170?

also, should be make two variables for the areas (firstRectangleArea, secondRectangleArea)? Am not sure, it just may be easier to read.

Why do we prefer the second element when they're equal? This seems to be a new condition, so there should be a logical reason for this (i.e. the last element is also the leafiest elemtn of the tree and thus the most relevant). Otherwise, I think I might prefer the first one, strictly based on the concept of stable ordering algorithms.

MichaelKetting (changes requested):

Remotion/Web/Development.WebTesting/ScreenshotCreation/BrowserContentLocators/EdgeBrowserContentLocator.cs
Stefan-Ilic: > aren't those two redundant and actually introduce more complexity given they are covered by the condition in line 170?

This statement would be correct for `System.Drawing.Rectangle`, but in the wonderful world of `System.Windows.Rect` the `Empty's` height and width are not `0` but `-Infinity`, making their area `Infinity` which is naturally larger than any other area, so we have to handle these special cases separately.

Since you had to ask, it might make sense to add a code comment for this peculiar situation.

And no, we cannot rely on their leafiness, I'll just return the first one if they're equal.

MichaelKetting (approved):

Remotion/Web/Development.WebTesting.IntegrationTests.Infrastructure/CustomWebTestConfigurationFactory.cs
```diff
@@ -96,6 +98,26 @@ protected override IBrowserConfiguration CreateCustomBrowserConfiguration (WebTe
...
+          .Where (x => x.StartsWith (browserName))
+          .ToDictionary (fileName => fileName, fileName => CreateVersionFromBrowserZipName (fileName, browserName))
+          .OrderByDescending (kvp => kvp.Value)
+          .First (kvp => kvp.Value.Major == int.Parse (latestTestedMajorBrowserVersion))
+          .Key;
+    }
+
+    private Version CreateVersionFromBrowserZipName (string fileName, string browserName)
+    {
+      var regex = new Regex (@$"^{browserName}_v(?<major>\d*).?(?<minor>\d*).?(?<patch>\d*).?(?<build>\d*)");
```
MichaelKetting, line 114: I'm thinking this should be easier/more compact. Can't we just say, there's either just a major version or a four-part-version and nothing in between? We can always pad with zeros on the right.

I don't know if this would be easier to read, but right now, there's a lot of code for very basic version parsing.

MichaelKetting: Also,  this whole version detection should not be part of the upgrade-Edge PR but rather a separate issue. Don't know if it needs to happen first, but it shoud not be the same PR.

MichaelKetting (changes requested):

Stefan-Ilic:
Added to `master` with 77a993673ebdd85e543b709b25034f5985db81de

PR #131, opened 2020-09-28
RM-7507: the postback sequence counter is being set too early on the client side to reliably detect the end of all client side changes

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7507

### DIFF
--- a/Remotion/Web/Core/ExecutionEngine/WxePageInfo.cs
+++ b/Remotion/Web/Core/ExecutionEngine/WxePageInfo.cs
@@ -54,7 +54,15 @@ namespace Remotion.Web.ExecutionEngine
 
     public static readonly string ReturningTokenID = "wxeReturningTokenField";
     public static readonly string PageTokenID = "wxePageTokenField";
+
+    /// <remarks>
+    /// Also used in Remotion/Web/Core/Res/HTML/ExecutionEngine.js
+    /// </remarks>
     public static readonly string PostBackSequenceNumberID = "wxePostBackSequenceNumberField";
+
+    /// <remarks>
+    /// Also used in Remotion/Web/Core/Res/HTML/ExecutionEngine.js
+    /// </remarks>
     public static readonly string DmaPostBackSequenceNumberID = "dmaWxePostBackSequenceNumberField";
 
     private const int HttpStatusCode_ServerError = 500;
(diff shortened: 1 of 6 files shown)


Remotion/Web/Core/Res/HTML/ExecutionEngine.js
```diff
@@ -70,7 +70,10 @@ function WxePage_Context(
     ArgumentUtility.CheckNotNullAndTypeIsBoolean('hasSubmitted', hasSubmitted);
     ArgumentUtility.CheckNotNullAndTypeIsBoolean('isCached', isCached);
 
-    document.getElementById('wxePostBackSequenceNumberField').setAttribute('value', _postBackSequenceNumber);
+    let dmaWxePostBackSequenceNumberField = document.getElementById('dmaWxePostBackSequenceNumberField');
+
+    if (dmaWxePostBackSequenceNumberField)
```
MichaelKetting, line 75: I typicall use an explict null-check in JS, see line 114. It makes the intent more clear when this is not just a sanity check but has semantics associated with it.

Remotion/Web/Core/ExecutionEngine/WxePageInfo.cs
```diff
@@ -288,7 +290,11 @@ public void OnPreRenderComplete ()
 
       _page.ClientScript.RegisterHiddenField (_page, WxeHandler.Parameters.WxeFunctionToken, wxeContext.FunctionToken);
       _page.ClientScript.RegisterHiddenField (_page, WxePageInfo.ReturningTokenID, null);
-      _page.ClientScript.RegisterHiddenField (_page, WxePageInfo.PostBackSequenceNumberID, wxeContext.PostBackID.ToString ());
+      int nextPostBackID = wxeContext.PostBackID + 1;
+      _page.ClientScript.RegisterHiddenField (_page, WxePageInfo.PostBackSequenceNumberID, nextPostBackID.ToString ());
+
+      if (SafeServiceLocator.Current.GetInstance<IRenderingFeatures>().EnableDiagnosticMetadata)
+        _page.ClientScript.RegisterHiddenField (_page, WxePageInfo.DmaPostBackSequenceNumberID, wxeContext.PostBackID.ToString());
```
MichaelKetting, line 297: make a second variable currentPostBackID remove code duplication and make it clear that these two are intentionally different usages of the postback ID. Otherwise, it would look like it's an oversight.

MichaelKetting (changes requested):

MichaelKetting (approved):

Remotion/Web/Core/Res/HTML/ExecutionEngine.js
```diff
@@ -64,6 +64,14 @@ function WxePage_Context(
   {
     ArgumentUtility.CheckNotNullAndTypeIsBoolean('hasSubmitted', hasSubmitted);
     ArgumentUtility.CheckNotNullAndTypeIsBoolean('isCached', isCached);
+
+    let dmaWxePostBackSequenceNumberField = document.getElementById('dmaWxePostBackSequenceNumberField');
+
+    if (dmaWxePostBackSequenceNumberField != null)
+    {
+      let postBackSequenceNumber = document.getElementById('wxePostBackSequenceNumberField').getAttribute('value');
```
MichaelKetting, line 72: Ah, I just realized how we do it in SmartPage:

* access to value is done directly: document.getElementById('wxePostBackSequenceNumberField').value instead of get/setAttribute.

* In SmartPage.js / SmartPageInfo (ca line 330), we pass the ID of the hidden field into the javascript's ctor. And if the field does not exist (i.e. the dma field during 'normal' operations), the ID itself is already set to null.

MichaelKetting (changes requested):

Remotion/Web/Core/Res/HTML/ExecutionEngine.js
```diff
@@ -59,11 +66,21 @@ function WxePage_Context(
...
+
   // Handles the page loading event.
   this.OnLoading = function(hasSubmitted, isCached)
   {
     ArgumentUtility.CheckNotNullAndTypeIsBoolean('hasSubmitted', hasSubmitted);
     ArgumentUtility.CheckNotNullAndTypeIsBoolean('isCached', isCached);
+
+    if (_dmaWxePostBackSequenceNumberFieldID != null)
+    {
+      let dmaWxePostBackSequenceNumberField = document.getElementById(_dmaWxePostBackSequenceNumberFieldID);
```
MichaelKetting, line 80: here it's let, above it's 'var'. pelase use var for now if the rest of the method also uses var and there's not benefit of let.

MichaelKetting (approved):
please fix the var/let thing and that's it.

iam-mholle:
Added to master with 40823c1a1d8258ca39856676fd5a5aa9b5385b9b.

PR #132, opened 2020-09-28
RM-7493: Add configurable timeouts to RetryUntilTimeout.Run methods

raffaelfoidl:
https://re-motion.atlassian.net/browse/RM-7493

### DIFF
--- a/Remotion/Web/Development.WebTesting/Utilities/RetryUntilTimeout.cs
+++ b/Remotion/Web/Development.WebTesting/Utilities/RetryUntilTimeout.cs
@@ -55,27 +55,27 @@ namespace Remotion.Web.Development.WebTesting.Utilities
       _retryUntilTimeout.Run();
     }
 
-    public static void Run ([NotNull] Action action, TimeSpan timeout = default, TimeSpan retryInterval = default)
+    public static void Run ([NotNull] Action action, TimeSpan? timeout = null, TimeSpan? retryInterval = null)
     {
       ArgumentUtility.CheckNotNull ("action", action);
       var configuration = new WebTestConfigurationFactory ().CreateDriverConfiguration();
 
       var retryUntilTimeout = new RetryUntilTimeout (
           action,
-          timeout == default ? configuration.SearchTimeout : timeout,
-          retryInterval == default ? configuration.RetryInterval : retryInterval);
+          timeout ?? configuration.SearchTimeout,
+          retryInterval ?? configuration.RetryInterval);
       retryUntilTimeout.Run();
     }
 
-    public static TReturnType Run<TReturnType> ([NotNull] Func<TReturnType> func, TimeSpan timeout = default, TimeSpan retryInterval = default)
+    public static TReturnType Run<TReturnType> ([NotNull] Func<TReturnType> func, TimeSpan? timeout = null, TimeSpan? retryInterval = null)
     {
       ArgumentUtility.CheckNotNull ("func", func);
       var configuration = new WebTestConfigurationFactory ().CreateDriverConfiguration();
 
       var retryUntilTimeout = new RetryUntilTimeout<TReturnType> (
           func,
-          timeout == default ? configuration.SearchTimeout : timeout,
-          retryInterval == default ? configuration.RetryInterval : retryInterval);
+          timeout ?? configuration.SearchTimeout,
+          retryInterval ?? configuration.RetryInterval);
       return retryUntilTimeout.Run();
     }
   }
--- a/Remotion/Web/Development.WebTesting/Utilities/RetryUntilTimeout.cs
+++ b/Remotion/Web/Development.WebTesting/Utilities/RetryUntilTimeout.cs
@@ -55,27 +55,27 @@ namespace Remotion.Web.Development.WebTesting.Utilities
       _retryUntilTimeout.Run();
     }
 
-    public static void Run ([NotNull] Action action)
+    public static void Run ([NotNull] Action action, TimeSpan timeout = default, TimeSpan retryInterval = default)
     {
       ArgumentUtility.CheckNotNull ("action", action);
       var configuration = new WebTestConfigurationFactory ().CreateDriverConfiguration();
 
       var retryUntilTimeout = new RetryUntilTimeout (
           action,
-          configuration.SearchTimeout,
-          configuration.RetryInterval);
+          timeout == default ? configuration.SearchTimeout : timeout,
+          retryInterval == default ? configuration.RetryInterval : retryInterval);
       retryUntilTimeout.Run();
     }
 
-    public static TReturnType Run<TReturnType> ([NotNull] Func<TReturnType> func)
+    public static TReturnType Run<TReturnType> ([NotNull] Func<TReturnType> func, TimeSpan timeout = default, TimeSpan retryInterval = default)
     {
       ArgumentUtility.CheckNotNull ("func", func);
       var configuration = new WebTestConfigurationFactory ().CreateDriverConfiguration();
 
       var retryUntilTimeout = new RetryUntilTimeout<TReturnType> (
           func,
-          configuration.SearchTimeout,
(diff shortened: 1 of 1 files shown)


Remotion/Web/Development.WebTesting/Utilities/RetryUntilTimeout.cs
```diff
@@ -55,27 +55,27 @@ public void Run ()
...
     {
       ArgumentUtility.CheckNotNull ("action", action);
       var configuration = new WebTestConfigurationFactory ().CreateDriverConfiguration();
 
       var retryUntilTimeout = new RetryUntilTimeout (
           action,
-          configuration.SearchTimeout,
-          configuration.RetryInterval);
+          timeout == default ? configuration.SearchTimeout : timeout,
+          retryInterval == default ? configuration.RetryInterval : retryInterval);
```
raffaelfoidl, line 66: I was hesitant to use the ternary conditional operator in this instance as the `configuration` variable is created in any case, even if both `timeout` and `retryInterval` are `default`. One possible alternative would be introducing a private method encapsulating the case distinction, something like this:
```c#
private Tuple<TimeSpan, TimeSpan> GetTimeoutAndRetryInterval(TimeSpan timeout, TimeSpan retryInterval)
    {
      if (timeout != default && timeout != default)
        return Tuple.Create (timeout, retryInterval);

      var configuration = new WebTestConfigurationFactory().CreateDriverConfiguration();
      return Tuple.Create (timeout == default ? configuration.SearchTimeout : timeout, retryInterval == default ? configuration.RetryInterval : retryInterval);
    }
```

Although, in my opinion, we would not gain much by doing so, except that things more would become more verbose than necessary. Hence, I went with the approach in the PR.

Remotion/Web/Development.WebTesting/Utilities/RetryUntilTimeout.cs
```diff
@@ -55,27 +55,27 @@ public void Run ()
       _retryUntilTimeout.Run();
     }
 
-    public static void Run ([NotNull] Action action)
+    public static void Run ([NotNull] Action action, TimeSpan timeout = default, TimeSpan retryInterval = default)
```
iam-mholle, line 58: The default of a TimeSpan is `00:00:00`, which is a valid timeout/interval. I would use `TimeSpan?` here and below.

Remotion/Web/Development.WebTesting/Utilities/RetryUntilTimeout.cs
iam-mholle: Using `TimeSpan?` as mentioned above makes this easier, you could just write `timeout ?? configuration.SearchTimeout` or even `timeout ??= configuration.SearchTimeout;` as separate statement.

Concerning the above, imo the ternary operator is a better choice here. It's just so much more code + tuples to extract it into a separate method. But that's just my opinion.

iam-mholle (changes requested):

Remotion/Web/Development.WebTesting/Utilities/RetryUntilTimeout.cs
raffaelfoidl: Yes, that makes sense. I guess I took the ticket's wording a bit too literally ("(...) should be configurable (...) via default `TimeSpan` parameters).

iam-mholle (approved):

Stefan-Ilic (approved):
Nice 👍

Remotion/Web/Development.WebTesting/Utilities/RetryUntilTimeout.cs
```diff
@@ -55,27 +55,27 @@ public void Run ()
       _retryUntilTimeout.Run();
     }
 
-    public static void Run ([NotNull] Action action)
+    public static void Run ([NotNull] Action action, TimeSpan? timeout = null, TimeSpan? retryInterval = null)
```
MichaelKetting, line 58: given the extra parameters, this now requires xmldoc. the doc for the parameters (or the remarks) must also explain, how the timeout and the retryinterval interact.

MichaelKetting (changes requested):

Remotion/Web/Development.WebTesting/Utilities/RetryUntilTimeout.cs
```diff
@@ -55,27 +56,58 @@ public void Run ()
       _retryUntilTimeout.Run();
     }
 
-    public static void Run ([NotNull] Action action)
+    /// <summary>
+    /// Executes a given <see cref="Action"/> repeatedly (using the given retry interval) until no exception is thrown during execution or until the
+    /// given timeout has been reached (in which case the final exception is rethrown).
```
MichaelKetting, line 61: last instead of final. Please check rest of new docs.

Remotion/Web/Development.WebTesting/Utilities/RetryUntilTimeout.cs
```diff
@@ -55,27 +56,58 @@ public void Run ()
...
     }
 
-    public static void Run ([NotNull] Action action)
+    /// <summary>
+    /// Executes a given <see cref="Action"/> repeatedly (using the given retry interval) until no exception is thrown during execution or until the
+    /// given timeout has been reached (in which case the final exception is rethrown).
+    /// </summary>
+    /// <param name="action">The <see cref="Action"/> to be executed repeatedly.</param>
+    /// <param name="timeout">
+    /// The timeout after which no more retries are made and the final exception is rethrown. Can be <see langword="null" />. If <see langword="null" />,
```
MichaelKetting, line 65: ...rethrown. Defaults to <see cref="DriverConfiguration.SearchTimeout"/> if no value is provided.

Remotion/Web/Development.WebTesting/Utilities/RetryUntilTimeout.cs
```diff
@@ -55,27 +56,58 @@ public void Run ()
...
+    /// the value of the <see cref="DriverConfiguration.SearchTimeout"/> property of the <see cref="DriverConfiguration"/> retrieved by calling
+    /// <see cref="WebTestConfigurationFactory.CreateDriverConfiguration"/> is assumed.
+    /// </param>
+    /// <param name="retryInterval">
+    /// The interval to wait between two executions. Can be <see langword="null" />. If <see langword="null" />, the value of the
```
MichaelKetting, line 70: see above

Remotion/Web/Development.WebTesting/Utilities/RetryUntilTimeout.cs
```diff
@@ -55,27 +56,58 @@ public void Run ()
...
           action,
-          configuration.SearchTimeout,
-          configuration.RetryInterval);
+          timeout ?? configuration.SearchTimeout,
+          retryInterval ?? configuration.RetryInterval);
       retryUntilTimeout.Run();
     }
 
-    public static TReturnType Run<TReturnType> ([NotNull] Func<TReturnType> func)
+    /// <summary>
```
MichaelKetting, line 86: see above

MichaelKetting (changes requested):

Remotion/Web/Development.WebTesting/Utilities/RetryUntilTimeout.cs
raffaelfoidl: I adopted the wording from the existing class documentation, which uses "final". Using "last" would introduce inconsistencies. Do you want me to change all such instances to "last" or keep it as it is?

MichaelKetting: I'm going with option c), please fix the xmldocs on the type to consistently use "last".

MichaelKetting (approved):

raffaelfoidl:
Added to master with e725fd1d02c39bc99f1a1e079d31b906994ddc2a.

PR #136, opened 2020-10-28
RM-7518: WxePostbackOutOfSequenceException is thrown when doubleclicking the pagination buttons of a BocList

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7518

### DIFF
--- a/Remotion/Web/Core/Res/HTML/SmartPage.js
+++ b/Remotion/Web/Core/Res/HTML/SmartPage.js
@@ -1167,6 +1167,7 @@ function SmartPage_Context(
     }
 
     var submitState = {
+      CancelSubmit: false,
       IsAsynchronous: isAsynchronous,
       IsAutoPostback: isAutoPostback,
       Submitter: submitterElement,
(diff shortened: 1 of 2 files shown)


Stefan-Ilic (approved):
I can confirm that the fix works for the repro provided. That's where my usefulness ends here, though

Remotion/Web/Core/Res/HTML/SmartPage.js
```diff
@@ -1178,6 +1178,8 @@ function SmartPage_Context(
       _submitState = submitState;
     else if (_submitState.IsAutoPostback && !isAutoPostback)
       _submitState.NextSubmitState = submitState;
+    else
+      _submitState.NextSubmitState = { CancelSubmit: true };
```
MichaelKetting, line 1182: I'm thinking, for consistency reasons, we should add CancelSubmit=false to line 1169 and Submitter=null to line 1182. This should make it more explicit that we have considered the relevant attributes and decided to make them null/false.

Please do, re-check the double postback catch behavior and the BocList bugfix, then we should be done.

MichaelKetting (changes requested):

Remotion/Web/Core/Res/HTML/SmartPage.js
iam-mholle: Tested it, works as intended.

MichaelKetting (approved):

iam-mholle:
Added to master with cce74420b42baba8598504163f93281e53637e90.

PR #143, opened 2020-11-16
RM-7530: Catch and log InvalidOperationException while pulling Docker image

raffaelfoidl:
https://re-motion.atlassian.net/browse/RM-7530

### DIFF
--- a/Remotion/Web/Development.WebTesting/HostingStrategies/DockerHosting/DockerClientException.cs
+++ b/Remotion/Web/Development.WebTesting/HostingStrategies/DockerHosting/DockerClientException.cs
@@ -1,0 +1,31 @@
+// This file is part of the re-motion Core Framework (www.re-motion.org)
+// Copyright (c) rubicon IT GmbH, www.rubicon.eu
+//
+// The re-motion Core Framework is free software; you can redistribute it
+// and/or modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2.1 of the
+// License, or (at your option) any later version.
+//
+// re-motion is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+
+namespace Remotion.Web.Development.WebTesting.HostingStrategies.DockerHosting
+{
+  /// <summary>
+  /// The exception that is thrown when a docker command fails.
+  /// </summary>
+  public class DockerClientException : Exception
+  {
+    public DockerClientException (string message)
+        : base (message)
+    {
+    }
+  }
+}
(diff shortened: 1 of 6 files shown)


iam-mholle (approved):

MichaelKetting (changes requested):
This design uses the IDockerClient like a 3rd party API with a pre-defiend contract. We should consider a other design-options first before catching any and all InvalidOperationExceptions thrown by this code. Please make suggestions :)

iam-mholle:
> 
> 
> This design uses the IDockerClient like a 3rd party API with a pre-defiend contract. We should consider a other design-options first before catching any and all InvalidOperationExceptions thrown by this code. Please make suggestions :)

Valid point with the `InvalidOperationException`, the client should throw a more precise exception that can be caught. Apart from that, I don't really see the issue, retrying a pull/ignoring the error is not part of the responsibility of the simple `docker pull` abstraction the method represents, that's something the user of it should do. What about a decorator ignoring the error the pull throws?

MichaelKetting:
> 
> 
> > This design uses the IDockerClient like a 3rd party API with a pre-defiend contract. We should consider a other design-options first before catching any and all InvalidOperationExceptions thrown by this code. Please make suggestions :)
> 
> Valid point with the `InvalidOperationException`, the client should throw a more precise exception that can be caught. Apart from that, I don't really see the issue, retrying a pull/ignoring the error is not part of the responsibility of the simple `docker pull` abstraction the method represents, that's something the user of it should do. What about a decorator ignoring the error the pull throws?

re decorator: perhaps. 
custom exception: yes, that's one way to go.

I was also considering a result (success/failed) or a parameter (fail quietly). Sometimes, I'm still unmanaged C :D But yes, a custom excpetion is probably okay for this.

Remotion/Web/Development.WebTesting/HostingStrategies/DockerHosting/DockerClientException.cs
```diff
@@ -0,0 +1,31 @@
...
+// along with re-motion; if not, see http://www.gnu.org/licenses.
+//
+using System;
+
+namespace Remotion.Web.Development.WebTesting.HostingStrategies.DockerHosting
+{
+  /// <summary>
+  /// The exception that is thrown when a docker command fails.
+  /// </summary>
+  public class DockerClientException : Exception
```
MichaelKetting, line 24: DockerOperationException, inlcude the result-code in the exception as separate field.

Remotion/Web/Development.WebTesting/HostingStrategies/IisDockerContainerWrapper.cs
```diff
@@ -49,7 +52,14 @@ public class IisDockerContainerWrapper : IDisposable
...
     public void Run ()
     {
-      _docker.Pull (_configurationParameters.DockerImageName);
+      try
+      {
+        _docker.Pull (_configurationParameters.DockerImageName);
+      }
+      catch (DockerClientException ex)
+      {
+        s_log.Error ("Pulling docker image failed", ex);
```
MichaelKetting, line 61: log the image name. 
end sentence with a .
log fallback behavior.

MichaelKetting (changes requested):

MichaelKetting (approved):

iam-mholle:
Added to master with 03eba1708ba00b7b24c8e1ae31129a21b8d292df.

PR #146, opened 2020-11-24
RM-7520: Clicking a button while the BocAutoCompleteReferenceValue suggestions are open cancels the button postback in Chrome

iam-mholle:
https://re-motion.atlassian.net/browse/RM-7520

### DIFF
--- a/Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js
+++ b/Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js
@@ -234,18 +234,18 @@
                     state.mouseDownOnSelect = false;
 
                     if (event.keyCode == KEY.RETURN) {
-                        var isValueSelected = selectCurrent();
+                        var isValueSelected = selectCurrent(false);
                         if (isValueSelected) {
                             //SelectCurrent already does everything that's needed.
                         } else {
                             var selectedItem = select.selected(true);
                             var isAnnotationSelected = selectedItem != null && selectedItem.data.IsAnnotation === true;
                             if (!isAnnotationSelected) {
-                                acceptCurrent(true);
+                                acceptCurrent(true, false);
                             }
                         }
                     } else {
-                        acceptCurrent(true);
+                        acceptCurrent(true, false);
                     }
 
                     if (event.keyCode == KEY.RETURN) {
@@ -353,15 +353,16 @@
             if (state.mouseDownOnSelect) {
                 informationPopUp.hide();
             } else {
+                var focusInputAfterSelection = false;
                 var isLastKeyPressBeforeBlurHandled = state.lastKeyPressCode == -1;
                 if (isLastKeyPressBeforeBlurHandled) {
-                    closeDropDownListAndSetValue($input.val());
+                    closeDropDownListAndSetValue($input.val(), focusInputAfterSelection);
                     updateResult ({ DisplayName : $input.val(), UniqueIdentifier : options.nullValue });
                 } else {
                     clearTimeout(timeout);
                     var lastKeyPressCode = state.lastKeyPressCode;
                     invalidateResult();
-                    acceptInput(lastKeyPressCode);
+                    acceptInput(lastKeyPressCode, focusInputAfterSelection);
                 }
             }
         }).click(function() {
@@ -432,7 +433,7 @@
                 event.stopPropagation();
 
                 if (select.visible()) {
-                    acceptInput (state.lastKeyPressCode);
+                    acceptInput (state.lastKeyPressCode, true);
                 } else {
                     $input.focus();
                     onChange(true, $input.val());
@@ -441,7 +442,7 @@
             });
         }
 
-        function acceptInput(lastKeyPressCode) {
+        function acceptInput(lastKeyPressCode, focusInputAfterSelection) {
             var isLastKeyPressedNavigationKey = false;
             switch (lastKeyPressCode) {
                 case KEY.UP:
@@ -464,17 +465,17 @@
                 select.selectItem (index);
             }
 
-            if (isLastKeyPressedNavigationKey && selectCurrent()) {
+            if (isLastKeyPressedNavigationKey && selectCurrent(focusInputAfterSelection)) {
                 //SelectCurrent already does everything that's needed.
             } else if (lastKeyPressCode != -1) {
-                acceptCurrent(true);
+                acceptCurrent(true, focusInputAfterSelection);
             } else {
-                closeDropDownListAndSetValue(state.previousValue);
+                closeDropDownListAndSetValue(state.previousValue, focusInputAfterSelection);
             }
         };
 
         // re-motion: allows empty input and invalid input
-        function acceptCurrent(confirmValue) {
+        function acceptCurrent(confirmValue, focusInputAfterSelection) {
             var term = $input.val();
             var selectedItem = null;
             if (confirmValue && term != '' && select.visible())
@@ -486,7 +487,7 @@
                 selectedItem = select.selected(false);
               }
             }
-            closeDropDownListAndSetValue(term);
+            closeDropDownListAndSetValue(term, focusInputAfterSelection);
 
             if (state.previousValue == term && selectedItem != null) {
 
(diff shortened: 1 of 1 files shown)


Stefan-Ilic (approved):
Looking good, I also tested the basic functionality of the control after the change an that seems to work as well, great job!

MichaelKetting (changes requested):
This fix may still leave race conditions. See follow-up issues on focus getting reset. therefor, maybe we can just skip the focus-restore if the user clicked outside the AutoComplete control.

Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js
```diff
@@ -1463,7 +1463,10 @@
                 field.selectionEnd = end;
             }
         }
-        field.focus();
+
+        if (focusAfterSelection && document.activeElement === document.getElementById(options.dropDownButtonID)) {
```
MichaelKetting, line 1467: this should be easier, not need to rely on activeELement. We know when the user triggered the dropdownbutton. this mens we know that the drop down button is the active element and therefor may set the focus to the field. There should also not be any addiitonal delay etc during which the user could click elsewhere while we process the first click, so this should alsways be straight forward. that the theory atleast.

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js
```diff
@@ -1445,7 +1445,7 @@
         };
     };
 
-    $.Autocompleter.Selection = function(field, start, end) {
+    $.Autocompleter.Selection = function(field, start, end, focusInputAfterSelection) {
```
MichaelKetting, line 1448: okay, that's strange. it doesn't look like this is ever called with false. is this correct?

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js
iam-mholle: `function autoFill(query, sValue) {` (l. 616) calls `Selection`:
```javascript
$.Autocompleter.Selection(input, query.length, query.length + sValue.length);
```

iam-mholle: Should false be passed in explicitly here? I thought it makes sense to leave the value to default here.

iam-mholle: Oh well, need to check one more time if this works in IE anyway.

MichaelKetting: no, please pass true/false everywhere. otherwise, it's hard to do a good review and later on read the code. Plus, undefined !== false in Javascript

iam-mholle: Alright, will do.

iam-mholle:
The current state of this PR does not fix the issue in Internet Explorer. Updated https://re-motion.atlassian.net/browse/RM-7520 accordingly.

Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js
```diff
@@ -1445,7 +1445,10 @@
         };
     };
 
-    $.Autocompleter.Selection = function(field, start, end) {
+    $.Autocompleter.Selection = function(field, start, end, focusInputAfterSelection) {
+        if (field !== document.activeElement && !focusInputAfterSelection)
+            return;
+
```
iam-mholle, line 1451: This condition would only be needed in 1.20 due to the existing IE support.

MichaelKetting: yeah, atm I'd just do IE. we will make the cut and cleanup in January when we setup a feature-branch-support for 1.20 in CI

Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js
```diff
@@ -354,13 +354,13 @@
             } else {
                 var isLastKeyPressBeforeBlurHandled = state.lastKeyPressCode == -1;
                 if (isLastKeyPressBeforeBlurHandled) {
-                    closeDropDownListAndSetValue($input.val());
+                    closeDropDownListAndSetValue($input.val(), false);
```
MichaelKetting, line 357: 💡We should introduce a {{focusInputAfterSelection}} varaible in line 355 to unify the state in lines 357 and 363.

Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js
```diff
@@ -1445,7 +1445,10 @@
         };
     };
 
-    $.Autocompleter.Selection = function(field, start, end) {
+    $.Autocompleter.Selection = function(field, start, end, focusInputAfterSelection) {
```
MichaelKetting, line 1448: It's funny, there is just one path that actually leads to the {{focusInputAfterSelection}} parameter being true ({{acceptInput}} when clicking the drop down button. And yet, due to the list of conditions until this point is reached, it doesn't seem like a good idea to just set the focus early on before doing all those nested calls.

MichaelKetting (changes requested):

Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js
```diff
@@ -1445,7 +1445,10 @@
         };
     };
 
-    $.Autocompleter.Selection = function(field, start, end) {
+    $.Autocompleter.Selection = function(field, start, end, focusInputAfterSelection) {
+        if (field !== document.activeElement && !focusInputAfterSelection)
```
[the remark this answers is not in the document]
iam-mholle: I could not find a case while thinking it through and testing it, do you have a guess if this could happen?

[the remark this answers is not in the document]
MichaelKetting: nope, which is why I tried to out-source it.

Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js
```diff
@@ -1044,7 +1045,7 @@
                 activeItem.addClass (CLASSES.ACTIVE);
                 activeItem.attr("aria-selected", "true");
 
-                select();
+                select(false);
```
MichaelKetting, line 1048: That's nice. (see comment in next line). Please create a ticket for this and we schedule it for after when this PR is done. We might be able to get rid of line 1050 and pass {{true}} instead. Will require detailed analysis, of course.

Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js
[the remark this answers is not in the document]
MichaelKetting: okay, it's now only in IE anyway and there is just one path where focus=true but the user is not currently typing and that's the dropdownbutton click. And there we already do the mousedownonselect, so, yes, we're done.

MichaelKetting (approved):

Remotion/ObjectBinding/Web/Res/HTML/BocAutoCompleteReferenceValue.jquery.js
iam-mholle: Added https://re-motion.atlassian.net/browse/RM-7546.

iam-mholle:
Added to master with 3f605fd906b9dd52e2cb3aae89e035e4c167084d.

